PR #18, opened 2017-11-20
Push notifications mods

Ivansss:
- [x] Add Notification Service Extension to modify push notifications
- [x] Decrypt push notifications
- [x] Modify push notification alert views

### DIFF
--- a/NotificationServiceExtension/NotificationService.m
+++ b/NotificationServiceExtension/NotificationService.m
@@ -1,0 +1,67 @@
+//
+//  NotificationService.m
+//  NotificationServiceExtension
+//
+//  Created by Ivan Sein on 14.11.17.
+//  Copyright © 2017 struktur AG. All rights reserved.
+//
+
+#import "NotificationService.h"
+
+#import "UICKeyChainStore.h"
+#import "NCSettingsController.h"
+
+@interface NotificationService ()
+
+@property (nonatomic, strong) void (^contentHandler)(UNNotificationContent *contentToDeliver);
+@property (nonatomic, strong) UNMutableNotificationContent *bestAttemptContent;
+
+@end
+
+@implementation NotificationService
+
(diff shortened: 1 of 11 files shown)


NotificationServiceExtension/NotificationService.m
```diff
@@ -0,0 +1,75 @@
...
+    }
+    
+    if (decryptedMessage) {
+        NSData *data = [decryptedMessage dataUsingEncoding:NSUTF8StringEncoding];
+        id messageJSON = [NSJSONSerialization JSONObjectWithData:data options:0 error:nil];
+        NSString *app = [messageJSON objectForKey:@"app"];
+        NSString *type = [messageJSON objectForKey:@"type"];
+        NSString *subject = [messageJSON objectForKey:@"subject"];
+        
+        if ([app isEqualToString:@"spreed"]) {
```
mario, line 49: spreed or talk?

NotificationServiceExtension/NotificationService.m
```diff
@@ -0,0 +1,75 @@
...
+            self.bestAttemptContent.title = @"";
```
mario, line 50: Title empty?

NotificationServiceExtension/NotificationService.m
Ivansss: The app id is still spreed.

NotificationServiceExtension/NotificationService.m
Ivansss: Yes :)

img_0094

Ivansss: @jancborchardt What do you think about these alerts?

mario: OK, and what about chat? Why do we show that when we don't support chat on mobile devices? :)

Ivansss: It will be supported at some point. (For now, we can show a message saying that it is not supported at the moment and provide and option to open web interface?)

mario:
I wouldnt show it.

PR #61, opened 2018-02-16
Fix numerical user ids

### LINKED ISSUE #60: App crash with users with numerical IDs
App crashed with users with numerical IDs.

Ivansss:
Fix #60


mario:
When can it be numerical

Ivansss:
For example, just try creating a user "123456"

mario:
Should still be a string xD

Ivansss:
I receive it as a long not as a string.

PR #147, opened 2018-08-31
Fix wording and typo on log in page

jancborchardt:
Please check @Ivansss! :)

It would also be good that when you press "OK" on the popup, the "Server address" field will automatically be focused again.


Ivansss (approved):

Ivansss:
Thanks @jancborchardt 😉 
Now the focus is returned to the input field after a failed login :)

jancborchardt:
@Ivansss good stuff, thanks! :)

Issue #115, opened 2018-06-11
No notifications


ghost:
Hello,
I don't have any notifications when a chat comes on IOS. Notification apps in Nextcloud is enabled and Notification is allowed for Nextcloud Talk in IOS.

Ivansss:
Hi @tomaaa ,
At the moment, there are only notifications for calls, invitations and chat mentions.

ghost:
Hi Ivansss,
thanks for your reply. I don't get any notifications for calls also.

Ivansss:
Ok... 🤔 
I read your post here: https://help.nextcloud.com/t/no-notifications-nextcloud-talk-ios-chat/32642
Did you try to log-out and log-in again?
Also, you could try to log-out, then remove all the sessions of the Nextcloud Talk app from the web UI (Settings->Security) and reinstall the iOS app.

Patate2:
I tried that.  and.... no, nothing, niet, nada.               
I'm crying.

mario:
Do you have the notifications app installed on the server?

Ivansss:
@Patate2 could you create a test account in your instance and send it to `ivan AT nextcloud DOT com`?
So I could debug what could be wrong in the subscription process?

mario:
The notifications are supposed to work out of the box as we use a Nextcloud-company-hosted service that sends the push messages through Google/Apple. 

All data sent is encrypted, as you can check in the notifications app.

jsalort:
@mario: OK, I understand. Thanks for the clarification. 

So, if I understand, you send a [POST request](https://github.com/nextcloud/talk-ios/blob/d03a6ef4b84bc3e020f7f4ba05bd0e8f245fcd79/VideoCalls/NCAPIController.m#L698) to [kNCPushServer](https://github.com/nextcloud/talk-ios/blob/d03a6ef4b84bc3e020f7f4ba05bd0e8f245fcd79/VideoCalls/NCSettingsController.m#L35) to subscribe the device to the Nextcloud-company-hosted notification service at push-notifications.nextcloud.com.
I guess the Notification app running on the instance already knows about this, and takes care of sending push notification to push-notifications.nextcloud.com (even if it is not clearly stated in the Notification app description that requests are going to be sent to a third-party server).

So, to track down where the notification gets lost, I suppose it should be possible to provide you with some log history:
- Is there a way to log into this nextcloud-hosted service, in order to verify that the registration was indeed successful ?
[...]

PR #152, opened 2018-09-12
Log error description on NC server notification subscription failure

jsalort:
I would like to have debugging information to figure out why the app can't subscribe to my instance push notifications.
See https://github.com/nextcloud/talk-ios/issues/115


Ivansss:
Closed due to #115 is fixed.

PR #184, opened 2018-12-14
New calling view design

Ivansss:
new-calling-ui

1. Waiting view
2. In call with 1 participant
3. In call with 2 participants (no custom avatar)


mario:
Can you please screenshot when you do design changes?

Ivansss:
Still WIP @mario. I was going to upload some screenshots when finished 😉

jancborchardt:
@Ivansss really nice! :) Only 1 thing: Now the circular avatar with blurry background and full-screen picture are mixed.

We should probably decide on one?
I'm fine with both, but in the issue where it was discussed, people seemed to prefer the one with circular avatar in the middle and blurry avatar as background.

The avatar circle (on your middle screenshot) should also be a bit larger, probably twice the size than it is now?

Ivansss:
@jancborchardt so the idea behind mixing both designs was to differentiate waiting view from an actual call.
While waiting you will see the blurry avatar as in the background (or background color of the default avatar) and when the connection is established with the other peer, then the avatar is shown (in case of a custom avatar the background is even more blurry as you can see above).
Here is an example with default avatar :)

default-avatar-call

jancborchardt:
@Ivansss right, good thinking! :slightly_smiling_face: Still, could we increase the size of the avatar/placeholder circle in the middle? For calls with more than 1 person it’s ok, but for 1on1 calls it could be twice the size. :)

Ivansss:
@jancborchardt sorry, I forgot to mention something about the avatar size on my last message 😅 
I also think the avatar for 1on1 calls could be bigger. 
I will play around a bit with the size to see which one fits better 😉

PR #209, opened 2019-03-04
Check app group restriction

Ivansss:
Check if user is allowed to use Nextcloud Talk due to group restrictions.
Related to https://github.com/nextcloud/spreed/pull/1585


mario:
What's the deal here? The user will simply not have a spreed capability, is that not right?

Ivansss:
> What's the deal here? The user will simply not have a spreed capability, is that not right?

Yes, that's it. Just adding some more user-friendly error dialogs :)

PR #265, opened 2019-11-13
Add the passcode to the application.

sconway856:
I followed the same code and approach from the nextcloud application:
	•	The passcode on and off is referred to as "lock" in the settings
	•	Simply passcode refers to entering 6 number. When "Weak passcode" in the settings is turned on (the default) then "Simply" mode is used.
	•	If the simply passcode is off and the "lock" is on, then a 64 alpha-numeric passcode is required.
	•	The passcode must be re-entered to "remove" the passcode.
	•	If the passcode is on and the weak setting is changed, thee original passcode must be entered followed by the new passcode and a confirmation of the new passcode.
	•	The mode of "simply" passcode is stored in the keychain "ncBlockCodeUseSimply"
	•	The passcode value is stored in the keychain value "ncBlockCode"
Here is a summary of my changes:
	1.	The code for the BKPasscodeView group and CCBKPasscode were copied from the nextcloud application.
	2.	In NCSettingsController I added the block code and "simply" passcode variables. (Not sure this is really necessary).
	3.	In the settings view controller is all the logic for adding, removing, and switching the weak passcode (aka "simply") on and off while the passcode is set.
	4.	Images were added for the dots on the simply screen and the switch for the weak password.
Please let me know if I should change anything in the patch file and I can re-send it.


Info.plist
```diff
@@ -0,0 +1,153 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sconway856: This can be reverted. It is only here for allowing http during my testing.

ThirdParty/BKPasscodeView/BKPasscodeDummyViewController.m
```diff
@@ -0,0 +1,41 @@
+//
```
sconway856: This whole group was copied from the ios nextcloud source codebase

VideoCalls.xcodeproj/project.pbxproj
```diff
@@ -1540,6 +1604,30 @@
 					"$(inherited)",
 					"$(PROJECT_DIR)/ThirdParty/openssl",
 				);
+				OTHER_LDFLAGS = (
```
sconway856: Revisit why this is a change.

ThirdParty/BKPasscodeView/BKTouchIDSwitchView.m
```diff
@@ -0,0 +1,125 @@
...
+}
+
+- (void)doneButtonPressed:(id)sender
+{
+    if ([self.delegate respondsToSelector:@selector(touchIDSwitchViewDidPressDoneButton:)]) {
+        [self.delegate touchIDSwitchViewDidPressDoneButton:self];
+    }
+}
+
+@end
```
sconway856: This ends the code copied from NextCloud/iOs

VideoCalls/CCBKPasscode.h
```diff
@@ -0,0 +1,49 @@
+//
```
sconway856, line 1: THis was also copied from nextcloud/ios

Ivansss:
Hi @sconway856,
After a quick test of your PR, I have noticed that I can't set a passcode.
Is that a known issue?

sconway856:
@Ivansss The passcode setting should work.  I am curious what you mean by "doesn't work".  You are able to open the application and get to the settings screen, but when you click on "lock" the passcode screen does not come up?  I am also curious if it is an update problem (if you uninstall and then install would it work)?

Ivansss:
https://github.com/nextcloud/talk-ios/pull/265/commits/1d17d2bb3737d3134a10d6b17dd599b0563a72c7 fixes the problem saving the passcode in physical iOS devices.
Now the feature works as expected in physical iOS devices and simulators.

Ivansss:
Maybe we could use a pod for the BKPasscodeView library instead of copying it.
What do you think @sconway856 ?
https://github.com/bkook/BKPasscodeView#podfile

VideoCalls/Info.plist
```diff
@@ -27,6 +27,11 @@
 	</array>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
+	<key>NSAppTransportSecurity</key>
+	<dict>
+		<key>NSAllowsArbitraryLoads</key>
+		<true/>
+	</dict>
```
Ivansss: I know you have used this for development :)
But we do not want to remove ATS from this project, so we should remove this code from this PR.

VideoCalls/NCUserInterfaceController.m
```diff
@@ -305,4 +305,5 @@ - (void)authenticationViewControllerDidFinish:(AuthenticationViewController *)vi
     }];
 }
 
+
```
Ivansss: We should remove also this change :D

Ivansss (changes requested):

VideoCalls/Info.plist
sconway856: Will do!

sconway856:
> Maybe we could use a pod for the BKPasscodeView library instead of copying it.
> What do you think @sconway856 ?
> https://github.com/bkook/BKPasscodeView#podfile

I like the idea.  i believe the first time I looked at it there were modifications to the BKPasscodeView in the nextcloud ios app.  I didn't look through them line by line, but I can take another look.  Would anyone on the nextcloud ios team know why they are not using the pod?

sconway856:
I will be out the next couple of day, but I will take a look when I get back on Sunday.

sconway856:
> I will be out the next couple of day, but I will take a look when I get back on Sunday.

Just finishing this up.  Looks like I can just use the library directly.

sconway856:
I update how the settings are displayed for lock on and for the weak passcode.  They now more closely match the video setting.  I think it now makes it more clear which state it is in.

The BKPasscodeView is now a pod file, and the extraneous changes are now all cleaned up and removed.

Ivansss:
Awesome work @sconway856 :)
Now the PR is much cleaner with the pod.
Also, the change in the settings looks and fits better with the rest of the settings.

Ivansss:
I changed the lock screen icon to the "lock" icon that we use in other parts of the app.
Also I added a new icon for privacy settings (before we were using the "lock" icon there).
What do you think @sconway856 ?

**Lock screen ON**
IMG_E0262

**Lock screen OFF**
IMG_E0263

sconway856:
> I changed the lock screen icon to the "lock" icon that we use in other parts of the app.
> Also I added a new icon for privacy settings (before we were using the "lock" icon there).
> What do you think @sconway856 ?
> 
> **Lock screen ON**
> IMG_E0262
> 
> **Lock screen OFF**
> IMG_E0263

Awesome!  I really like these changes, thanks!

Ivansss:
I added and icon for the password type, so every setting has an icon now.
Also changed "Weak" to "Simple".

IMG_E0264

Ivansss (approved):

Ivansss:
Thanks for your awesome work @sconway856! :)
This feature will be included in the next release.

Ivansss:
Hi @sconway856,
We found some issues with this feature when using FaceID while testing the app before the v8.0 release.
Sadly we had to disable this feature for now until we find a fix for it.

jancborchardt:
Hey @sconway856, we recently established a "Talk development chat" for all contributors. We can add you there with a guest account via your email if you like – would be cool to have you! :)

PR #303, opened 2020-04-28
Handle the new reference id for snappier chat feeling

Ivansss:
A temporary message is added to the chat right after the 'Send' button is tapped.
This temporary message is replaced with the one coming from the server if the sending succeeded.

out1

If the message sending failed. An error indicator is shown next to failed message.

fail

Rel: https://github.com/nextcloud/spreed/pull/3077


jancborchardt:
Very very nice! :) Possible enhancements:
- The spinner could look simpler, like the one in our server
- There could be some slight animation of the message up from the input field, I think iMessage does this?

Ivansss:
> The spinner could look simpler, like the one in our server

Yeah. I also thought the same. I will try to add the one we use in the server in a following PR :)

> There could be some slight animation of the message up from the input field, I think iMessage does this?

That could be a nice enhancement, but not an easy one. I will put it in the backlog.
When there are more messages (enough to cover the whole screen), there is already an animation. The new message will push up all the other messages.

PR #376, opened 2020-10-15
l10n: Change video to camera

Valdnet:


NextcloudTalk/CallViewController.m
```diff
@@ -739,7 +739,7 @@ - (IBAction)videoCallButtonPressed:(id)sender
 - (void)showUpgradeToVideoCallDialog
 {
     UIAlertController *confirmDialog =
-    [UIAlertController alertControllerWithTitle:NSLocalizedString(@"Do you want to enable your video?", nil)
+    [UIAlertController alertControllerWithTitle:NSLocalizedString(@"Do you want to enable your camera?", nil)
                                         message:NSLocalizedString(@"If you enable your video, this call will be interrupted for a few seconds.", nil)
```
Ivansss: Maybe we should also change it here:
If you enable your video -> If you enable your camera

Ivansss (changes requested):

Ivansss (approved):

NextcloudTalk/CallViewController.m
Valdnet: Oops. Overlooked 😉.

PR #377, opened 2020-10-15
Correct punctuation in CallParticipantViewCell.m

Valdnet:


jancborchardt (approved):
Looks good, yes indeed we put a space between the last word and … :)

Issue #393, opened 2020-11-01
Prevent starting a call before joining the room completed


SystemKeeper:
So I've been debugging some issues which happened from time to time:
* After ending a call you're dropped back to the room list instead of the chat
* You're unable to enter a room (nothing happens when you select the row)
* You start a video call and receive "Unable to join call"

#### How to reproduce
* Open talk app (this is with external signaling, but I think it does not matter)
* Enter chat with `user1`
* Now quicky do 3 things
  * Open notification center
  * Close notification center
  * Immediatly start a video call and wait for it to connect
* Leave video call
* Notice that you're dropped back to room list instead of the chat with `user1`
* Try to enter chat with `user1` again -> nothing happens
* Enter chat with `user2` and leave chat with `user2` again
* Enter chat with `user1` and start a video call again
* "Could not join call"
* In some cases leaving or joining the conversation with `user1` will now not send join/leave events anymore

Because we're haven't joined the conversation with user1 the api returns 404 when trying to start a call. This leads to "Could not join call".

#### Cause
When the app becomes inactive (as happens when opening the notification center), `appWillResignActive` in `NCChatViewController.m` will fire and call `leaveChatInRoom` in `NCRoomsManager`.

After resuming (closing the notification center), `appDidBecomeActive` fires and then calls `joinRoom` in `NCRoomsManager`. As joining is async, we still can start a video call, even without finishing the join (at least that's what it looks like). When now leaving the call, we're dropped back to the rooms list and unable to join the previous room again.

#### Resolution
I was unable to fix the root cause of this. As a workaround I made sure that the call-controls are disabled until joining the room completed.
@Ivansss I guess this problem could also occur when receiving a call via push notification and CallKit and could then lead to a "Could not join call" message, but I was unable to test this. Maybe you have a idea how this could be fixed properly ?

Ivansss:
Fixed in #394

PR #394, opened 2020-11-01
Prevent starting a call when room wasn't joined yet

SystemKeeper:
This happens after app becomes active again ("appDidBecomeActive") but joining the room did not finish yet.

Ref #393


Ivansss (approved):

Ivansss:
Nice catch ;)

PR #398, opened 2020-11-03
Add push-to-talk functionality

SystemKeeper:
Currently talk-ios uses a dialog to display a warning when participant is force-muted by a moderator. In contrast nextcloud talk web uses a toast to display the warning, which will disappear after 7s (default timeout for toasts in nextcloud).

This PR changes the behaviour in talk-ios to also use a toast which will disappear automatically but can be dismissed by the user as well.

Before

image

After

image

### DIFF
--- a/NextcloudTalk/CallViewController.m
+++ b/NextcloudTalk/CallViewController.m
@@ -591,16 +591,19 @@ typedef NS_ENUM(NSInteger, CallState) {
         [self->_audioMuteButton setImage:[UIImage imageNamed:@"audio-off"] forState:UIControlStateNormal];
         
         NSString *micDisabledString = NSLocalizedString(@"Microphone disabled", nil);
+        NSTimeInterval duration = 1.5;
         UIView *toast;
         
         if (reason) {
+            // Nextcloud uses a default timeout of 7s for toasts
+            duration = 7.0;
+
             toast = [self.view toastViewForMessage:reason title:micDisabledString image:nil style:nil];
         } else {
             toast = [self.view toastViewForMessage:micDisabledString title:nil image:nil style:nil];
         }
         
-        // Nextcloud uses a default timeout of 7s for toasts
-        [self.view showToast:toast duration:7.0 position:CSToastPositionCenter completion:nil];
+        [self.view showToast:toast duration:duration position:CSToastPositionCenter completion:nil];
     });
 }
 
--- a/NextcloudTalk/CallViewController.m
+++ b/NextcloudTalk/CallViewController.m
@@ -50,7 +50,7 @@ typedef NS_ENUM(NSInteger, CallState) {
     CallStateInCall
 };
 
-@interface CallViewController () <NCCallControllerDelegate, UICollectionViewDelegate, UICollectionViewDelegateFlowLayout, UICollectionViewDataSource, RTCEAGLVideoViewDelegate, CallParticipantViewCellDelegate>
+@interface CallViewController () <NCCallControllerDelegate, UICollectionViewDelegate, UICollectionViewDelegateFlowLayout, UICollectionViewDataSource, RTCEAGLVideoViewDelegate, CallParticipantViewCellDelegate, UIGestureRecognizerDelegate>
 {
     CallState _callState;
     NSMutableArray *_peersInCall;
@@ -69,6 +69,8 @@ typedef NS_ENUM(NSInteger, CallState) {
     BOOL _videoCallUpgrade;
     BOOL _hangingUp;
     PulsingHaloLayer *_halo;
+    PulsingHaloLayer *_haloPushToTalk;
+    UIImpactFeedbackGenerator *_buttonFeedbackGenerator;
 }
 
 @property (nonatomic, strong) IBOutlet UIView *buttonsContainerView;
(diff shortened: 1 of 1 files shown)


jancborchardt:
> use a toast which will disappear automatically but can be dismissed by the user as well.

@SystemKeeper I am not so familiar with iOS (sometimes using a device but don’t have my own), is this toast a common pattern in iOS? As far as I remember I haven’t seen anything of the sort so far, or do you have a reference? :)

Maybe it would make more sense to use a nonmodal popover for this: https://developer.apple.com/design/human-interface-guidelines/ios/views/popovers/
It could show the text, be dismissed when tapping anywhere except the popover, and it could point to your own mute button, indicating the status.

And I assume this would appear in 2 cases:
- When you join the call initially
- When you tap on your own microphone icon

Ivansss:
@jancborchardt I think the toast is fine. The one we are using is not really iOS native, but iOS 14 is now showing some toasts for example when you paste something from the clipbloard.

We are using the toast that @SystemKeeper used here for notifying about things like "message copied", "mic muted", "camera disabled"...

The only thing I'm not really sure here is if the toast is good for warning about that a moderator muted you. Maybe you could miss the toast if you are not looking at the screen. Instead you can't miss it if we show a dialog.

What do you think @SystemKeeper @jancborchardt ?

jancborchardt:
> The only thing I'm not really sure here is if the toast is good for warning about that a moderator muted you. Maybe you could miss the toast if you are not looking at the screen. Instead you can't miss it if we show a dialog.

That’s why I was proposing the popover. :) It behaves like a dialog in that you need to tap somewhere to dismiss it, but also we can have it point to the mute button, which makes people aware of the context, and gives feedback as to why pressing the audio on/off button does not work.

E.g. if you are muted by a moderator and you try to unmute, this popover would also show, originating from the button you just tapped. :)

SystemKeeper:
@jancborchardt Those toasts are already used in the ios-talk app:

image

Just not for being force-muted. That’s why I adapted it this way. WhatsApp is using something similar:

image

@Ivansss Personally I think a toast would be enough: In my case participants are often muted, because everyone should be muted. In that case a notification which does not require user interaction would be enough and it‘s same in Web, the message disappears automatically.

I guess in a follow up commit we could do this more inline with iOS 14 and display the toast on top of the screen and maybe show a little „muted icon“ in a corner? This way it’s always clear to the user that he’s muted, regardless if he did see the toast or not.

Ivansss:
@jancborchardt That popovers are usually used in iPad interfaces. They even say "Avoid displaying popovers on iPhones." below the image in the link you shared https://developer.apple.com/design/human-interface-guidelines/ios/views/popovers/

SystemKeeper:
> E.g. if you are muted by a moderator and you try to unmute, this popover would also show, originating from the button you just tapped. :)

Hm, just to make sure I fully get this: When you’re force muted and you unmute yourself you want to display „Microphone enabled“ in the same popup, Right ?

SystemKeeper:
@jancborchardt For reference, this is what iOS 14 does when changing the ringer state:
image

Quick implementation of a permanent mute-indicator:

image
image
image

> [...] and gives feedback as to why pressing the audio on/off button does not work.

Maybe I'm missing something, but getting muted by a moderator does not mean I can't unmute myself, does it? So there would be no other toast/popover/... needed, because the button always works?!

Let me know what you think :-)

SystemKeeper:
Thinking about it, we could also display the (un-)mute button, in case we’re muted. This way it’s always clear to the user and one can immediately unmute himself. Not sure if this is good UX however 🧐

Ivansss:
> Maybe I'm missing something, but getting muted by a moderator does not mean I can't unmute myself, does it? So there would be no other toast/popover/... needed, because the button always works?!

You are right. Even if you have been muted by a moderator, you can unmute yourself.

I like the idea of the mute indicator, but this will only work in video calls and not in voice calls :/

My only concern here is if we should really make users aware that a moderator has muted them (so we show a permanent dialog as before) or if it is good enough to just show a toast for 7secs (even if users can miss it if they are not looking at the screen).

SystemKeeper:
> I like the idea of the mute indicator, but this will only work in video calls and not in voice calls :/

In voice calls the UI Elements are always visible, so there’s already an indicator via the mute button.

Thinking about the idea with permanently showing the mute button: this would also be a requirement for a future push-to-talk implementation, wouldn’t it? Short touch: unmute, press and hold: push-to-talk?

Regarding the toast, I still would argue it should be consistent in web and mobile, therefore a 7s toast would make sense. But I‘m fine either way :)

SystemKeeper:
> Thinking about the idea with permanently showing the mute button: this would also be a requirement for a future push-to-talk implementation, wouldn’t it? Short touch: unmute, press and hold: push-to-talk?

Quick implementation:

Force-muted:
image

While muted:
image

When controls show:
image

Button shows when force muted and stays there (also when you mute yourself of course). Long press could then be used for push-to-talk I guess?

SystemKeeper:
Latest commit permanently shows the unmute-button if you're muted. It also adds push-to-talk: Press-and-hold the unmute button to temporarily unmute yourself. You're muted again if you release the button.

Ivansss (approved):
Tested and works!

Ivansss:
Awesome work @SystemKeeper :)

PR #399, opened 2020-11-04
Fixes when joining a call while in a chat from another conversation

Ivansss:
This PR fixes 2 issues that occur when joining a call while being in a chat from another conversation.

1. The chat could not be enter after the call was picked or declined. Only restarting the app would allow to join that chat again.

2. If the app was in the background with an open chat and the call was picked, the call stayed in "Waiting for participants..." screen.


SystemKeeper (approved):
Looks good!

PR #403, opened 2020-11-06
Add "Go to file" to file-conversations

SystemKeeper:
Currently there's no way to open the file which belongs to a file-conversation. This PR adds a "Go to file" action to the room info which opens the file in nextcloud app.

**ToDo**
-  Figure out how we can get the file path with just the fileID.  To open a file in the iOS Nextcloud App, we need the internal link (done) and the path of the file. I've yet to figure out how to retrieve the path. Files which are shared in a room automatically have a path attribute, doesn't look like there's something like this for a file-room
- Check if it's ok to show the image of the filetype or if we need a dedicated icon here

image

image

### DIFF
--- a/NextcloudTalk/FileMessageTableViewCell.m
+++ b/NextcloudTalk/FileMessageTableViewCell.m
@@ -23,9 +23,7 @@
 #import "FileMessageTableViewCell.h"
 #import "SLKUIConstants.h"
 #import "MaterialActivityIndicator.h"
-#import "NCSettingsController.h"
 #import "NCUtils.h"
-#import "OpenInFirefoxControllerObjC.h"
 #import "UIImageView+AFNetworking.h"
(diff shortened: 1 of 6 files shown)


SystemKeeper:
> * Check if it's ok to show the image of the filetype or if we need a dedicated icon here

@jancborchardt  Do you have any recommendation on this one? :-)

SystemKeeper:
> * Figure out how we can get the file path with just the fileID.  To open a file in the iOS Nextcloud App, we need the internal link (done) and the path of the file. I've yet to figure out how to retrieve the path. Files which are shared in a room automatically have a path attribute, doesn't look like there's something like this for a file-room

Resolved, thanks to @Ivansss !

Ivansss (approved):
Tested and works 👍

Ivansss:
Awesome work once again ;)

PR #414, opened 2020-11-14
Add interactive push notifications

SystemKeeper:
This PR adds interactive push-notifications. When receiving a push-notification of type `chat` the user has the ability to directly reply to the message - directly from the notification:

image
image

I've done testing with my own push-implementation, so this needs testing with the NC-proxy, but should work out-of-the-box.


Ivansss (approved):
Tested and works 👍

PR #416, opened 2020-11-15
Use native context menus

SystemKeeper:
Since iOS 13 there's a native implementation of contextmenus available: https://developer.apple.com/design/human-interface-guidelines/ios/controls/context-menus/

If running on iOS 13 or later, this PR uses native contextmenus:

image
image

This is based on #415

### DIFF
--- a/NextcloudTalk/NCChatViewController.h
+++ b/NextcloudTalk/NCChatViewController.h
@@ -24,6 +24,8 @@
 
 #import "NCRoom.h"
 
+extern NSString * const NCChatViewControllerJoinChatAndReplyPrivately;
+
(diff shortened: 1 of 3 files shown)


SystemKeeper:
The last commit adds a custom previewProvider. This mainly has 2 advantages: 

1. The contextmenu is always beneath the preview (not above as seen in first screenshot) and therefore in line with other apps (mail for example). 
2. When the contextmenu for a groupedMessage is shown, it's displayed as a non-grouped-message. See screenshots:

image
image

Ivansss (approved):
Tested and works 👍

PR #417, opened 2020-11-16
Fix for leaving call action.

Ivansss:
Fixes ocassional app crashes when leaving a call.


SystemKeeper (approved):
I'm unable to reproduce the crash after this 👍

PR #418, opened 2020-11-16
Return best attempt content when notification service extension time expires

Ivansss:
Instead of returning a default notification body, now we return the best attempt we had.
So for example, if the notification service extension time expires before getting the notification from the server, we can return at least the decrypted notification content.


SystemKeeper (approved):
Tested and works 👍

PR #419, opened 2020-11-17
Add sharing support from camera

SystemKeeper:
Access camera from talk and share the photo directly.


Ivansss (approved):
Tested and works 👍

PR #422, opened 2020-11-19
Context menu fixes

Ivansss:
- [x] Do not show context menu in mentions tableview
- [x] Set correct size for latge messages and in landscape orientation
- [x] Fix iPad previews


SystemKeeper (approved):
👍

PR #430, opened 2020-11-28
Increase font-size for pure emoji-messages with max. 3 emojis

### LINKED ISSUE #220: Improvement: make single emojis bigger
Hello,

I have a suggestion. We could make messages bigger when they contain only emojis.

On macOS or iOS, if there is one, two or three emojis in a message there are bigger and not on a blue background.
On the contrary, if the message contains text or more than 3 emojis, it looks like a classic message.
Capture d’écran 2019-04-21 à 11 42 52

I think it would be better on Talk if a similar idea was implemented.
Currently it looks like this:
Capture d’écran 2019-04-21 à 11 43 56

SystemKeeper:
Fixes #220 

image


Ivansss (approved):
Tested and works 👍

PR #434, opened 2020-11-30
Share improvements

SystemKeeper:
This PR changes the following things:

* Allow uploading up to 5 images/files from ShareExtension (not from within talk currently)
* Generate preview of files on iOS >= 13
* Allow editing of images/files on iOS >= 13
* Allow editing text when sharing a text or url before send

image

image

image


Ivansss (approved):
Tested and works 👍 
Awesome improvements!! :)

PR #437, opened 2020-12-06
Download and view shared files in talk-ios

SystemKeeper:
This PR implements directly downloading and viewing files in talk-ios.

**Caching**
* A cache directory is created per account
* Files older than 7 days are automatically deleted
* Server is checked for newer versions of a file prior to using the cache
* Cache directory is removed upon account removal

**General**
* ActivityIndicator shows download progress
* File will be opened after download if 
  * The room wasn't left in between download start and finish
  * The corresponding cell which the file belongs to is still visible (not scrolled away)
* Not implemented for room-files currently, should be done in a follow-up
* Added "Open in Nextcloud" to context-menu


Ivansss (approved):
Tested and works! 👍
Awesome PR! 🎉

PR #438, opened 2020-12-06
Explicitly disable left-/rightButton when room is not joined

SystemKeeper:
When a room is not joined (or we're in offline mode) the share/send button is still active, although touch interaction is disabled:
image

This PR disables both buttons in connection with touch interaction. 
(The send button could be shown as "active" if there's an unsend message in the textview)


SystemKeeper:
Thinking about this... When we enable/disable the buttons, do we still need to disable touch interactions with the whole inputbar? With pending messages, someone could write a message even in offline mode for later sending. Also it wouldn’t prevent opening the keyboard if someone is faster touching the input bar than the room join..?!

Ivansss (approved):

PR #440, opened 2020-12-11
Fix missing icons in chat-view on iOS 10

### LINKED ISSUE #439: Anyone else missing Phone and Camera Icons from conversations?
Did an ios-Talk upgrade recently and the phone and Camera icons are missing from the top bar in conversations.
Can click the blank space and it starts a call, same with video chat.

10.1.2 on iPhone 6

SystemKeeper:
Fixes #439


Ivansss (approved):
Tested and works 👍

PR #444, opened 2020-12-13
Enable rightButton only when there's a message to send

### LINKED ISSUE #442: rightButton is enabled even without message
RightButton gets enabled even if there's no message in the inputbar, allowing to send empty messages.

Ref https://github.com/nextcloud/talk-ios/commit/3e1cb817a782edadba8cf9b8914822efde7ea0a6

SystemKeeper:
Fixes #442


Ivansss (approved):
Tested and works 👍

PR #466, opened 2021-01-12
Check type of capabilities-dictionary before save

SystemKeeper:
Otherwise: 
image
image
image


SystemKeeper:
Maybe `meta` statuscode should be checked NCAPIController in the future as well?

Ivansss (approved):

PR #477, opened 2021-01-21
Fix replyMessageView hiding the last message

SystemKeeper:
When quoting a message, which is not the last message, you end up i a state, where the replyMessageView is above the last received message. Also when quoting older messages, sometimes rows are not fully rendered.

This also solves some animation quirks.


SystemKeeper:
Still not happy... will take another look

SystemKeeper:
Should work fine now

Ivansss (approved):

PR #478, opened 2021-01-22
Allow notifications to be dismissed

SystemKeeper:
When a notification is displayed to the user, he can dismiss it either by opening the notification and tapping the "X"-button or by swipping left and selecting "Clear". Until now this removes the notification from notification center, but doesn't change the bade on talk-ios. Now the badge is decreased when actively dismissing a notification.


Ivansss:
As discussed, we prefer to keep the app icon badge when a user manually removes a notification from the notification center.
Thanks for bringing up the topic :)

PR #492, opened 2021-02-14
Periodically check if we accepted the call on another device

### LINKED ISSUE #433: Ringtones keeps ringing
Hello, is there a way to have the app detect once you join the call from a different devices. So what i can see happing is if someone calls me all of my phones will ring but if I answer it from my computer or one of the phones the the other phones will still keep ringing then eventually it will stop . Is there a way to have the app detect that you've answered the call and stop ringing?

SystemKeeper:
Fixes #433 

When receiving an incoming call we periodically check if we're already in that call (eg. when we answered the call on a second device).

Two things I'm wondering:

https://github.com/nextcloud/talk-ios/blob/ef1515fd465a2523581ed222d9ec82cc8de2a279/NextcloudTalk/CallKitManager.m#L316-L322

1. Could we check here if **any** participant has the `inCall` flag set and assume if **none** has the `inCall` flag, the call ended before we accepted it?

https://github.com/nextcloud/talk-ios/blob/ef1515fd465a2523581ed222d9ec82cc8de2a279/NextcloudTalk/CallKitManager.m#L325-L326

2. Is the `weakSelf` required here?

### DIFF
--- a/NextcloudTalk/CallKitManager.h
+++ b/NextcloudTalk/CallKitManager.h
@@ -37,6 +37,7 @@ extern NSString * const CallKitManagerWantsToUpgradeToVideoCall;
 @property (nonatomic, strong) NSString *accountId;
 @property (nonatomic, strong) CXCallUpdate *update;
 @property (nonatomic, assign) BOOL reportedWhileInCall;
+@property (nonatomic, assign) BOOL isRinging;
 
 @end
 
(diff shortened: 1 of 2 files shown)


SystemKeeper:
https://github.com/nextcloud/talk-ios/blob/ef1515fd465a2523581ed222d9ec82cc8de2a279/NextcloudTalk/CallKitManager.m#L317-L321

Just noticed, should this be weakSelf as well?

Ivansss (approved):

PR #502, opened 2021-02-22
Differentiate audio and video calls in incoming calls

Ivansss:
- Differentiate audio or video calls in incoming call screen.
- Answer video calls with video disabled by default.

### DIFF
--- a/NextcloudTalk/CallKitManager.m
+++ b/NextcloudTalk/CallKitManager.m
@@ -410,6 +410,7 @@ NSString * const CallKitManagerWantsToUpgradeToVideoCall        = @"CallKitManag
     if (call) {
         [self stopHangUpTimerForCallUUID:call.uuid];
         NSMutableDictionary *userInfo = [NSMutableDictionary dictionaryWithObject:call.token forKey:@"roomToken"];
+        [userInfo setValue:@(call.update.hasVideo) forKey:@"hasVideo"];
         [userInfo setValue:@(call.reportedWhileInCall) forKey:@"waitForCallEnd"];
         [[NSNotificationCenter defaultCenter] postNotificationName:CallKitManagerDidAnswerCallNotification
                                                             object:self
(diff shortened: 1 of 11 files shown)


NextcloudTalk/CallKitManager.m
```diff
@@ -235,15 +235,19 @@ - (void)getCallInfoForCall:(CallKitCall *)call
...
-        }];
     }
+    
+    TalkAccount *account = [[NCDatabaseManager sharedInstance] talkAccountForAccountId:call.accountId];
+    [[NCAPIController sharedInstance] getRoomForAccount:account withToken:call.token withCompletionBlock:^(NSDictionary *roomDict, NSError *error) {
+        if (!error) {
+            NCRoom *room = [NCRoom roomWithDictionary:roomDict andAccountId:call.accountId];
+            [self updateCall:call withDisplayName:room.displayName];
+            
+            NSInteger callFlag = [[roomDict objectForKey:@"callFlag"] integerValue];
```
SystemKeeper, line 246: No capability-check needed here?

Ivansss: No. If the "callFlag" attribute is not there, callFlag will be 0. So it will notify an audio-only call as in the past when no "callFlag" attribute was sent.

SystemKeeper: Okay, wasn't sure if you would get an NSInteger for a nil value.

Ivansss: No problem ;) Thanks a lot for taking a look at the code :)
But yeah, [nil integeValue] will return 0.

nickvergessen: Lazy. 0 could mean everyone left already again. We added a capability for this so you could trust the value:
https://nextcloud-talk.readthedocs.io/en/latest/capabilities/#111

But on iOS you have to present a call screen anyway, right? Even if the other person already left again

Ivansss: > But on iOS you have to present a call screen anyway, right? Even if the other person already left again

Yes. That's why I can only differentiate between presenting and audio or a video incoming call :D

SystemKeeper: But we could add this in https://github.com/nextcloud/talk-ios/pull/492 to end the call when the other person already left.

SystemKeeper: Hm, but you could end the call immediatly if the capability is there and it's 0, right? At this point the call was already reported and we can hang up again.

Ivansss: You are right! We could add it in a following PR ;)

Ivansss: Implemented in #507

PR #507, opened 2021-02-26
Check "conversation-call-flags" capability 

Ivansss:
Check "conversation-call-flags" capability and hang up the call if the call has ended already.

### DIFF
--- a/NextcloudTalk/CallKitManager.m
+++ b/NextcloudTalk/CallKitManager.m
@@ -246,6 +246,7 @@ NSString * const CallKitManagerWantsToUpgradeToVideoCall        = @"CallKitManag
             if ([[NCSettingsController sharedInstance] serverHasTalkCapability:kCapabilityCallFlags forAccountId:call.accountId]) {
                 NSInteger callFlag = [[roomDict objectForKey:@"callFlag"] integerValue];
                 if (callFlag == CallFlagDisconnected) {
+                    [self presentMissedCallNotificationForCall:call];
                     [self endCallWithUUID:call.uuid];
                 } else if ((callFlag & CallFlagWithVideo) != 0) {
                     [self updateCall:call hasVideo:YES];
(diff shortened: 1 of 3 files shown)


SystemKeeper:
When the call is ended after 45s from the hangup timer we still show a notification about the missed call. I’m not sure if we should do the same here?
Asking the other way round, in which cases should we add such notification?

NextcloudTalk/CallKitManager.m
```diff
@@ -243,9 +243,14 @@ - (void)getCallInfoForCall:(CallKitCall *)call
...
             [self updateCall:call withDisplayName:room.displayName];
             
-            NSInteger callFlag = [[roomDict objectForKey:@"callFlag"] integerValue];
-            BOOL hasVideo = (callFlag >= CallFlagWithVideo);
-            [self updateCall:call hasVideo:hasVideo];
+            if ([[NCSettingsController sharedInstance] serverHasTalkCapability:kCapabilityCallFlags forAccountId:call.accountId]) {
+                NSInteger callFlag = [[roomDict objectForKey:@"callFlag"] integerValue];
+                if (callFlag == CallFlagDisconnected) {
+                    [self endCallWithUUID:call.uuid];
+                } else if (callFlag >= CallFlagWithVideo) {
```
SystemKeeper: Can we assume, that when `CallFlagWithPhone` is set it always means there is video? Because that would currently apply to this check (video is 4, phone is 8).
Otherwise I think it should be something like `(callFlag & CallFlagWithVideo) != 0` ?

Ivansss:
> When the call is ended after 45s from the hangup timer we still show a notification about the missed call. I’m not sure if we should do the same here?
> Asking the other way round, in which cases should we add such notification?

I think we should present it in both cases. Added missed call notification in 5894679af99e65a1c69652f8b2bcd405806acc4c

NextcloudTalk/CallKitManager.m
Ivansss: You are right ;) Added bitwise comparison here 6408e08f78e9ff5471dd5268103055df9d687dd4

SystemKeeper (approved):
LGTM!

PR #513, opened 2021-03-03
Call flags related fixes

Ivansss:
- Fixes participant's call flag indicator in room info
- Sends correct call flag when joining room

### DIFF
--- a/NextcloudTalk/NCRoomParticipants.m
+++ b/NextcloudTalk/NCRoomParticipants.m
@@ -22,6 +22,8 @@
 
 #import "NCRoomParticipant.h"
 
+#import "NCCallController.h"
+
 @implementation NCRoomParticipant
 
 + (instancetype)participantWithDictionary:(NSDictionary *)participantDict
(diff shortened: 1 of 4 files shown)


NextcloudTalk/NCRoomParticipants.m
```diff
@@ -72,4 +74,22 @@ - (NSString *)displayName
     }
     return _displayName;
 }
+
+- (NSString *)callIconImageName
+{
+    if (self.inCall == CallFlagDisconnected) {
```
SystemKeeper, line 80: Maybe we should rename `inCall` to `callFlag` to avoid confusion? Maybe better add a `isInCall` as boolean with a getter if we need it?

Ivansss: `inCall` is the name of the attribute that comes from the server (when fetching participants). Naming it `callFlag` would make sense but I think it could created confusion :P

https://github.com/nextcloud/talk-ios/blob/master/NextcloudTalk/NCRoomParticipants.m#L37

SystemKeeper: Oh, sorry, I thought in the backend it was callFlags, but it’s only called callFlag for conversations...

Ivansss: No problem ;) we kept the attribute name for legacy reasons. In the past it was a just boolen variable but then it was extend to include callFlags.

PR #515, opened 2021-03-05
Set a different cookie storage for each user.

### LINKED ISSUE #508: No notifications about messages  from iOS app. ver 11.0.0
No notifications about messages  from iOS app. since ver 11.0.0
it's the same issues that in june 1918!!!!
this problem is the return...
https://help.nextcloud.com/t/no-notifications-nextcloud-talk-ios-chat/32642

### LINKED ISSUE #5299: Can't leave room as $sessionId is null
I was investigating a report about missing notifications in talk-ios (https://github.com/nextcloud/talk-ios/issues/508), although the issue was resolved by the user, I was able to see the same on my NC21,  Talk 11.1 instance.

At the point of my tests, no browser sessions were open, just used talk-ios. I noticed, that the notification was not send, because there was still an active session:
https://github.com/nextcloud/spreed/blob/eb3de685e567aeed2193313402e79cc99c91584d/lib/Chat/Notifier.php#L384-L387

I then looked into talk-ios and the requests, but the DELETE to `/ocs/v2.php/apps/spreed/api/v3/room/{token}/participants/active` was successfully send and also received by the webserver, but the session was still in `oc_talk_sessions`.

After digging around, I found that the `leaveRoom` function was unable to retrieve my session (`$sessionId` was `null`):
https://github.com/nextcloud/spreed/blob/c642f3b889d9d73918945d6aea75b6999c3dbc77/lib/Controller/RoomController.php#L1672-L1685

This only happened with talk-ios to me, not with talk-web. Before splitting Attendees and Session the participant was resolved through the userId, if available:

https://github.com/nextcloud/spreed/blob/32f14e15c4b9330412a12b148b5554f6dd592f9c/lib/Controller/RoomController.php#L1281-L1300

If I use `$participant = $room->getParticipant($this->userId);` it works for me in my test cases.

After reading the comment at https://github.com/nextcloud/spreed/pull/5194#discussion_r585522533 I'm thinking that the TalkSession for the talk-ios request can't be determined, because the request was made on another PHP session? Could there be a bug in leaveRoom?

Ivansss:
In this PR we set a different cookie storage for each user. Also we now send cookies on every api request.

This fixes the problem of missing push notifications after leaving a conversation.
The app was not sending the session cookie in leaveRoom(), so the user was not notified on new messages in that conversation until the session was removed by the server after a timeout.

Fixes #508
Fixes https://github.com/nextcloud/spreed/issues/5299

### DIFF
--- a/NextcloudTalk/NCAPIController.h
+++ b/NextcloudTalk/NCAPIController.h
@@ -90,7 +90,6 @@ extern NSInteger const kReceivedChatMessagesLimit;
 @interface NCAPIController : NSObject
 
 @property (nonatomic, strong) NSMutableDictionary *apiSessionManagers;
-@property (nonatomic, strong) NSMutableDictionary *apiUsingCookiesSessionManagers;
 @property (nonatomic, strong) AFImageDownloader *imageDownloader;
 @property (nonatomic, strong) AFImageDownloader *imageDownloaderNoCache;
 
--- a/NextcloudTalk/NCAPIController.m
+++ b/NextcloudTalk/NCAPIController.m
@@ -71,7 +71,6 @@ NSInteger const kReceivedChatMessagesLimit = 100;
     _defaultAPISessionManager = [[NCAPISessionManager alloc] initWithSessionConfiguration:configuration];
     
     _apiSessionManagers = [NSMutableDictionary new];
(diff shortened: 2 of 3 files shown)


SystemKeeper (approved):
Works in my tests 👍

PR #517, opened 2021-03-07
Add background fetch

### LINKED ISSUE #516: Loading history can invalidate indexPath of unreadMessageSeparator
* Open chat with unread messages
* Load history by going to the top of the chat
* Write something and try to send it

-> Because of loading the history / adding other sections the saved indexPath of the unreadMessageSeparator can get invalid and we try to remove an invalid row.

SystemKeeper:
This PR adds the ability to fetch new messages in the background when enabled (Settings -> General -> Background App Refresh). 

Some remarks:
* When a background fetch occurs is up to iOS (when in low power mode background fetch might be disabled by the OS)
* When a background fetch occurs in normal operations, we update the rooms and update each chat for rooms that changed since the last sync (so viewing new messages should be instantaneous now)
* When rooms need to be updates while the app is in foreground (eg. update control, recevied push, ...) we do the same thing and directly load the new messages
* When in low power mode and a background fetch occurs, we only update the rooms, not the chats
* This can be tested in Xcode via Debug -> Simulate background fetch (note that even if you use a device >= iOS 13 this will trigger `performFetchWithCompletionHandler` instead of `handleAppRefresh`

I really don't get why this code was neccessary before https://github.com/nextcloud/talk-ios/commit/13ef2c73eaf9688fe2f3e8a5c92500974f013d18. We could re-introduce it again if there's a reason, but there need to be changes because we now add the unreadMessagesSeparator before the first request from the API is done.

Fixes #516

### DIFF
--- a/NextcloudTalk/AppDelegate.m
+++ b/NextcloudTalk/AppDelegate.m
@@ -32,6 +32,10 @@
 
 #import <UserNotifications/UserNotifications.h>
 
+#import <BackgroundTasks/BGTaskScheduler.h>
+#import <BackgroundTasks/BGTaskRequest.h>
+#import <BackgroundTasks/BGTask.h>
+
 #import "NCAudioController.h"
 #import "NCAppBranding.h"
 #import "NCConnectionController.h"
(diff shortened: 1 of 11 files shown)


Ivansss (approved):
Awesome work @SystemKeeper 🎉

PR #521, opened 2021-03-08
Improve error messages when connection to server fails

SystemKeeper:
Improve error message when connection to server fails on login. 
I noticed that not all error messages are translated by apple (for example `NSURLErrorAppTransportSecurityRequiresSecureConnection` is not). Should we add them to translate or keep as "OS-native" as possible here?

Ref https://help.nextcloud.com/t/nextcloud-talk-funktioniert-unter-ios-nicht/109430


Ivansss (approved):

Ivansss:
> I noticed that not all error messages are translated by apple (for example NSURLErrorAppTransportSecurityRequiresSecureConnection is not). Should we add them to translate or keep as "OS-native" as possible here?

I would keep it as is is. Apple might add a localizedDescription at some point.
And even if they don't, this already gives the user much more info than before :)

SystemKeeper:
Sounds good to me 👍

PR #526, opened 2021-03-14
Add intent/suggestion support

SystemKeeper:
This PR adds support for Siri-Suggestions, when enabled and iOS >= 13 is used. When sharing something, most recent talk conversation will be added as suggestions:
image

I'm not too happy with the avatar part (group and public avatars):
https://github.com/nextcloud/talk-ios/compare/suggestion-support?expand=1#diff-90e4fe91c9e8f643694eabd5d3c6611caf7fae9d79d7ed4f3e61870dd4f1e014R131-R150
But it does work fine in my test.


SystemKeeper:
For testing you can enable most recent suggestions in the developer menu.

Ivansss (approved):

PR #529, opened 2021-03-19
👤 Profile info

Ivansss:
Allow to view and edit profile information.


SystemKeeper:
Your tapRecognizer prevents selecting addAccount / removeAccount. Only long press works for me here.
Other than that personally I think the textFields should be in "disabled mode" (text is gray) when editing is not enabled, because they look editable right away.

PR #558, opened 2021-05-14
⚙️ API V4 support

Ivansss:


SystemKeeper:
Very nice!
Can we re-enable background-fetch for v4 then? https://github.com/nextcloud/talk-ios/blob/eb271ae27530f0a11f382521925c8896e59a96c2/NextcloudTalk/NCRoomsManager.m#L313-L328

SystemKeeper:
> Can we re-enable background-fetch for v4 then?

@Ivansss Any reason not to enable this now with v4?

PR #585, opened 2021-06-30
Fix Custom Signaling server url

gjongenelen:
The Custom Signaling server URL in the issue-template is invalid because of the duplicate `/index.php` prefix.


Ivansss (approved):

Ivansss:
Thanks @gjongenelen for noticing the typo and fixing it :)

PR #594, opened 2021-07-14
Fix chat notifications delay

### LINKED ISSUE #580: Push notification delay
I have a self hosted server, with ntp enabled and time zone set to utc, but the push notification seems to be delayed like about 30seconds, any reason for this.   Thanks

Ivansss:
Fix #580


DwinAnderson:
Thank you, @Ivansss! That really helped! No delays in notificatiions. If there are lots of messages, they go one after another. Cool fix!

PR #619, opened 2021-09-16
i18n: Correct spelling

Valdnet:


Ivansss:
Thanks for the PR @Valdnet :)
One question, how did you modify the `NextcloudTalk/en.lproj/Localizable.strings` file?
For some reason the resulting file size is half as the original one.

SystemKeeper:
It's converted to UTF-8 instead of using UTF-16 as before. 
Transifex expects (?) UTF-16: https://docs.transifex.com/formats/apple-strings
According to apple, UTF-8 is recommended and will be automatically converted to UTF-16 when the product is build: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/LoadingResources/Strings/Strings.html

UTF-8 probably would have the advantage of showing diffs on github, instead of treating it as a binary file. Not sure how transifex handles this change...

:man_shrugging:

Valdnet:
> One question, how did you modify the NextcloudTalk/en.lproj/Localizable.strings file?

@Ivansss I modified it directly in the GitHub repository.

If it is incorrectly modified, I am asking you to modify the file again 😉.

@SystemKeeper Thanks for the comprehensive information 😉.

Ivansss:
Thanks for the research and the explanation @SystemKeeper 😊

> UTF-8 probably would have the advantage of showing diffs on github, instead of treating it as a binary file. Not sure how transifex handles this change...

I'm not sure either :S

> If it is incorrectly modified, I am asking you to modify the file again 😉.

I will just convert it back to UTF-16 just in case, since the Apple tool [genstrings](https://github.com/nextcloud/talk-ios/blob/master/docs/Localization/Generate-localization-file.md#using-genstrings-tool) that we use to generate that file creates a UTF-16 file. Thanks again for the PR @Valdnet 😉

Ivansss (approved):

SystemKeeper:
You're welcome!
Just in case you want to change it, this would be a way: https://stackoverflow.com/a/69174370

PR #626, opened 2021-09-28
i18n: Spelling unification

Valdnet:
Spelling unification in Transifex.


rakekniven (approved):
@Ivansss  Why does GitHub think that "Localizable.strings" is binary?

Valdnet:
@rakekniven It is explained [here](https://github.com/nextcloud/talk-ios/pull/619#issuecomment-921058662) 😉.

Ivansss:
Did this string change in the server repo as well? @Valdnet @rakekniven

Valdnet:
@Ivansss To avoid recoding again, it was upload as a file.

Valdnet:
@Ivansss How can I remove a conflict? When solving it, it does not indicate anything.

rakekniven (changes requested):
Something went wrong when solving conflict of Localizable.strings using the web UI.
Please do not merge.
File in branch is empty now.

rakekniven (approved):

nickvergessen:
> Did this string change in the server repo as well?

yes

Ivansss:
Thanks for the PR @Valdnet and @rakekniven :)

SystemKeeper:
Maybe the best option would be to just edit the source file, instead of the `localizable.strings`, because the file is generated periodically by @Ivansss : https://github.com/nextcloud/talk-ios/blob/master/docs/Localization/Generate-localization-file.md ? Would make PRs and merging easier.

Ivansss:
I think what @SystemKeeper suggested would be the best idea for future PRs @Valdnet and @rakekniven 
You can just edit source files and after merging the PR I will use the [genstrings tool ](https://github.com/nextcloud/talk-ios/blob/master/docs/Localization/Generate-localization-file.md) to update `localizable.strings`

Valdnet:
@Ivansss Thanks a lot 👍. This will make it easier for us to edit the file with the text strings.

rakekniven:
You handle it different then Nextcloud ios app, or?

There I often only change https://github.com/nextcloud/ios/blob/develop/iOSClient/Supporting%20Files/en.lproj/Localizable.strings to make source string adaptions.

Ivansss:
Thanks for pointing to that file @rakekniven :)

I just noticed that https://github.com/nextcloud/ios/blob/develop/iOSClient/Supporting%20Files/en.lproj/Localizable.strings is UTF-8 encoded, and that's the file that is sent to transifex. So it seems that transifex also accepts UTF-8.
We were just not sure while discussing it [here](https://github.com/nextcloud/talk-ios/pull/619#issuecomment-921058662).

So we could convert https://github.com/nextcloud/talk-ios/blob/master/NextcloudTalk/en.lproj/Localizable.strings to UTF-8 and that way it could be easily modified (and also we would be able to see the changes in the diff).

Ivansss:
[en.lproj/Localizable.strings](https://github.com/nextcloud/talk-ios/blob/master/NextcloudTalk/en.lproj/Localizable.strings) is now UTF-8 encoded.
Now the changes in that file are shown nicely in the diff https://github.com/nextcloud/talk-ios/commit/5b7a6456785e4d66c55526f57f084c71542e4613 :)
Also the sync with transifex works as expected!

PR #631, opened 2021-10-01
Use XCFrameworks to support dev environments on apple silicon macs

SystemKeeper:
Currently the project can't be build on Macs using apples new M1 processors. This PR removes the legacy framework builds to use XCFrameworks to support M1 Macs.

* Need latest version of CocoaPods and Carthage (I used 1.11 and 0.38, see changes to README)
* Need at least XCode 12
* Removed the old `*.framework` reference and added the new `*.xcframework` reference
* Removed `carthage.sh` which was introduced as a workaround before using XCFrameworks
* Removed `copy-frameworks`-build phase according to carthage migration doc (see [here](https://github.com/Carthage/Carthage/#migrating-a-project-from-framework-bundles-to-xcframeworks))
* Exclude `arm64` platform for iOS simulator (as seen [here](https://stackoverflow.com/a/63955114/2512312))

- [x] Tested with iOS simulator and real device on an M1 Mac
- [ ] Tested with iOS simulator and real device on an *non-M1* Mac (@Ivansss I guess you can test this?)


SystemKeeper:
Closed in favor of SPM (#641 )

PR #633, opened 2021-10-02
Blur video of peer when the video stream is paused/frozen

### LINKED ISSUE #431: Enable Pause Theme Alert
Currently if a user opens another app while still being on talk app the audio works but video freezes. It would be nice to have a theme to state video paused. Current that's how it works while using WhatsApp. And FaceTime will work the same way if you turn off PIP feature. 

Below image is from FaceTime. 

image

The Talk app just freezes the picture at the last position. It's more virtual appeal 🙂

SystemKeeper:
Fixes #431 

When a peer with enabled video did not send video frames for a few seconds the view gets blurred like it is with other apps.

image

### DIFF
--- a/NextcloudTalk/NCPeerConnection.m
+++ b/NextcloudTalk/NCPeerConnection.m
@@ -30,16 +30,20 @@
 #import <WebRTC/RTCMediaStream.h>
 #import <WebRTC/RTCDefaultVideoEncoderFactory.h>
 #import <WebRTC/RTCDefaultVideoDecoderFactory.h>
+#import <WebRTC/RTCLegacyStatsReport.h>
 
 #import "ARDSDPUtils.h"
 
 #import "NCSignalingMessage.h"
 
+NSTimeInterval const kNCPeerConnectionCheckVideoPausedStateEverySeconds  = 3.0;
+NSTimeInterval const kNCPeerConnectionCheckVideoPausedStateWhilePausedEverySeconds  = 1.0;
 
 @interface NCPeerConnection () <RTCPeerConnectionDelegate, RTCDataChannelDelegate>
 
 @property (nonatomic, strong) RTCPeerConnectionFactory *peerConnectionFactory;
 @property (nonatomic, strong) NSMutableArray *queuedRemoteCandidates;
+@property (nonatomic, assign) NSInteger videoBytesReceivedByPeer;
 
 @end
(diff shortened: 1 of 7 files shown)


AndyXheli (approved):

NextcloudTalk/NCPeerConnection.m
```diff
@@ -30,16 +30,20 @@
 #import <WebRTC/RTCMediaStream.h>
 #import <WebRTC/RTCDefaultVideoEncoderFactory.h>
 #import <WebRTC/RTCDefaultVideoDecoderFactory.h>
+#import <WebRTC/RTCLegacyStatsReport.h>
 
 #import "ARDSDPUtils.h"
 
 #import "NCSignalingMessage.h"
 
+NSTimeInterval const kNCPeerConnectionCheckVideoPausedStateEverySeconds  = 3.0;
```
SystemKeeper, line 39: Not sure if 3s is to short for this, increase to 5s ?

SystemKeeper:
As this solution is only working in the iOS client, I'm wondering if it would make more sense to send a `videoOff` message when switching to another app?!
`[self sendDataChannelMessageToAllOfType:@"videoOff" withPayload:nil];`
Then all clients would would react as if the video was disabled, so most likely show the avatar.

As this PR also covers cases were no data is received, the `videoOff` could be done additionally (maybe even as a first step).

AndyXheli:
@SystemKeeper i think that would be ideal if someone wants to switch to a different app then blurs the screen

SystemKeeper:
@Ivansss as #822 is merged: Should we keep this around (it still might make sense, for example when on a bad connection), but on the other hand might increase complexity unnecessarily..?!
If we want to add this some time, I'd say we should increase the timer significantly (not 3s).

Ivansss:
I think #822 fixes nicely the problem of disabled video when the app is sent to the background.
We could use the code in this PR  to notifiy about connection problems with a participant (as the Web app does), but I would not use it for the initial purpose of this PR (detecting a disabled video) since we have now #822.
What do you think @SystemKeeper ? :)

SystemKeeper:
Yes, I agree, the original purpose of this PR does not make sense anymore. I like the idea about connectivity problems. I'll close this PR and reference it in a new issue, so we have this tracked.

PR #639, opened 2021-10-10
Make headers clickable

alazarevic5:
Signed-off-by: Aleksandra Lazarevic <aleksandra@nextcloud.com>


Ivansss (approved):
Tested and works 👍

PR #641, opened 2021-10-13
Add NCCommunication library as Swift Package

Ivansss:


SystemKeeper:
## General

* `README`  needs to be updated, to remove Carthage installation instructions and prerequisites
* `Framework Search Path` still contains `$(PROJECT_DIR)/Carthage/Build/iOS` on all 3 targets

## Tests

The following tests are made on macOS 11.6 (Big Sur) with Apple M1 processor. Each test is done with a freshly cloned repo and cleaned build folder. The following is used:

```bash
git clone https://github.com/nextcloud/talk-ios.git && \
cd talk-ios && \
git checkout communication-lib-spm && \
git submodule update --init && \
pod install && \
open NextcloudTalk.xcworkspace
```

### XCode with Rosetta

XCode is started with rosetta emulation (in `Finder -> XCode -> Get Info -> Open with Rosetta`)

#### Simulator

✔  NextcloudTalk target running on simulated iPhone 11 Pro with iOS 15.0

#### Real device

✔  NextcloudTalk target running on iPhone 11 Pro with iOS 15.0,2

### XCode native

XCode is started without rosetta emulation

#### Simulator

❌  NextcloudTalk target running on simulated iPhone 11 Pro with iOS 15.0

This mainly happens because some libraries we're using are distributed as prebuild libs with iOS target. I found the following problems:

* OpenSSL (libcrypto.a and libssl.a are in the repo)
* WebRTC

We can work around this by adding both as SPM modules as well, in my tests I added the following repos:

* https://github.com/krzyzanowskim/OpenSSL
* https://github.com/alexpiezo/WebRTC

(There might be others of course https://swiftpackageregistry.com/search?term=openssl)

Then I can run on simulator. This change is way out of scope for this PR, so I'll just leave it here for a follow-up PR and tests somewhere in the future.

#### Real device

✔  NextcloudTalk target running on iPhone 11 Pro with iOS 15.0,2


## Conclusion

This PR looks good and makes sense. It doesn't solve all problems with building on M1 Macs, but in my opinion is a future-proof way to add the NCCommunication lib to the project. It might be worth thinking about dropping Carthage support in NCComm-Lib as well, consindering that the dependencies are already out of sync compared to SPM.

I suggest we close #631 in favor of this one, but fix the 2 issues (README and Search Path) before.

SystemKeeper (approved):
👍

Ivansss:
Fixed remaining 2 issues @SystemKeeper ;)

What about the pod changes you did in #631?
https://github.com/nextcloud/talk-ios/pull/631/files#diff-11f9f23e03ad0b3f7f8706bf06c020b232b195eb43756d35fe22c9de4cbe4019R2146
https://github.com/nextcloud/talk-ios/pull/631/files#diff-11f9f23e03ad0b3f7f8706bf06c020b232b195eb43756d35fe22c9de4cbe4019R2196
https://github.com/nextcloud/talk-ios/pull/631/files#diff-8f7d6adf31268a2d897ee34bd170592648d6e520aa237104395e4a4438af50cbR53
Are those needed to build the project in a M1 Mac?

SystemKeeper:
I can build the project as per the comment above. Those changes might be neccessary when we want to build for simulator without rosetta. But I think we should have a look at this, when we replace OpenSSL and WebRTC to feature multiple architectures. All other 3 variants work without those changes, so I think we're good to go :)

PR #649, opened 2021-10-18
Conversation settings enhancements

Ivansss:
Display conversation settings as in the Web UI

IMG_0099 IMG_0100


SystemKeeper:
In Web "Leave Conversation" is not displayed red, should this be inline with web?
image

Maybe a different icon for both Delete-Options might make sense? 
https://fonts.google.com/icons?icon.query=delete
(I can see "Delete Forever" work for "Delete Conversation" or "Delete Sweep" for "Delete all messages")

Ivansss:
> In Web "Leave Conversation" is not displayed red, should this be inline with web?

I also thought about that. The thing is that "Leave conversation" is a destructive action if the conversation is not an "open conversation". So we either add an explanation (which will be a bit long for mobiles) or we leave it red for not opened coversations and gray for open ones.

> Maybe a different icon for both Delete-Options might make sense?
> https://fonts.google.com/icons?icon.query=delete
> (I can see "Delete Forever" work for "Delete Conversation" or "Delete Sweep" for "Delete all messages")

I also don't really like to have same icons next to each other for 2 different actions.
The icons you suggested I guess they would look good :)

What do you think @marcoambrosini @jancborchardt ?

NextcloudTalk/RoomInfoTableViewController.m
```diff
@@ -1644,19 +1598,19 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
...
         {
-            NSArray *actions = [self getPublicActions];
-            PublicAction action = [[actions objectAtIndex:indexPath.row] intValue];
+            NSArray *actions = [self getGuestsActions];
+            GuestAction action = [[actions objectAtIndex:indexPath.row] intValue];
             switch (action) {
-                case kPublicActionPublicToggle:
+                case kGuestActionPublicToggle:
                 {
                     UITableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:shareLinkCellIdentifier];
```
SystemKeeper: Rename to `allowGuestsCellIdentifier` or similar ?

NextcloudTalk/RoomInfoTableViewController.m
```diff
@@ -1536,65 +1516,39 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
...
+        case kRoomInfoSectionNotifications:
         {
-            NSArray *actions = [self getRoomActions];
-            RoomAction action = [[actions objectAtIndex:indexPath.row] intValue];
+            NSArray *actions = [self getNotificationsActions];
+            NotificationAction action = [[actions objectAtIndex:indexPath.row] intValue];
             switch (action) {
-                case kRoomActionFavorite:
-                {
-                    UITableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:favoriteRoomCellIdentifier];
```
SystemKeeper, line 1546: `favoriteRoomCellIdentifier` is still defined

SystemKeeper: Also `kModificationErrorFavorite` is there as well.

NextcloudTalk/RoomInfoTableViewController.m
Ivansss: Fixed!

NextcloudTalk/RoomInfoTableViewController.m
Ivansss: Fixed!

Ivansss:
This is the final look for the destructive actions

IMG_E0102

marcoambrosini:
I would go non-red with the explanation like in the web view for the leave conversation. I wouldn't worry too much about space as our scrolling muscles are pretty well trained nowadays and those are not very frequent actions anyway :) 

As per the action icons, do you have something similar to this for the delete messages? https://materialdesignicons.com/icon/delete-clock

Ivansss:
> As per the action icons, do you have something similar to this for the delete messages? https://materialdesignicons.com/icon/delete-clock

Do you mean for "delete all messages" action or for delete a single message in the chat view?

Ivansss:
Added a new icon for "Delete all messages" icon
https://materialdesignicons.com/icon/chat-remove-outline
Thanks @SystemKeeper for the finding :)

IMG_E0103

Also added information about "Leave conversation" in the confirmation dialog

IMG_E0104

marcoambrosini:
Looks good!

PR #651, opened 2021-10-20
Ignore .DS_Store files

SystemKeeper:


Ivansss (approved):
I have it in my git global settings but it makes sense to have it here as well just in case ;)

PR #654, opened 2021-10-26
Avatar hovercard

Ivansss:
iPhone

IMG_E0106  

iPad

IMG_E0026 

Fix #650

PR #656, opened 2021-10-28
Setup custom URL to open conversations from another app.

Ivansss:
Custom URL:
nextcloudtalk://open-conversation?server={serverURL}&user={userId}&withUser={userId}

### DIFF
--- a/NextcloudTalk/AppDelegate.m
+++ b/NextcloudTalk/AppDelegate.m
@@ -144,6 +144,21 @@
     // Called when the application is about to terminate. Save data if appropriate. See also applicationDidEnterBackground:.
 }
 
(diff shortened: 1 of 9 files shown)


SystemKeeper:
Is this „just“ a way to open a specific conversation through the link or is there any feature from talk which interacts with that?

Ivansss:
It is a way to open a conversation from another app or link.
The Files iOS app will use it to open a conversation after tapping on an avatar and selecting "Talk to user" option.
https://github.com/nextcloud/ios/issues/1715
We will also have these options when tapping on an avatar in the Talk iOS app #654

alazarevic5 (approved):
I tested with sermo. App has been opened with url like in the example.
But it redirected me to view controller with all convesations.

PR #657, opened 2021-10-29
Reduce number of assets

Ivansss:


alazarevic5 (approved):
All assets are okay, the images are displayed properly

PR #660, opened 2021-11-12
Show user status icon in conversations list 

alazarevic5:
Show user status icon in conversations list like in the web Talk app.
When status is everything instead do not disturb and somebody set custom status with emoji  (for e.g. working 💻) , emoji will be shown. In other cases status will be shown (online, away, dnd)
Screenshot 2021-11-12 at 15 02 30
Screenshot 2021-11-12 at 15 01 55
Screenshot 2021-11-12 at 15 01 35
.

### DIFF
--- a/NextcloudTalk/NCDatabaseManager.m
+++ b/NextcloudTalk/NCDatabaseManager.m
@@ -89,11 +89,6 @@ NSString * const kMinimumRequiredTalkCapability     = kCapabilitySystemMessages;
         configuration.schemaVersion = kTalkDatabaseSchemaVersion;
         configuration.migrationBlock = ^(RLMMigration *migration, uint64_t oldSchemaVersion) {
             // At the very minimum we need to update the version with an empty block to indicate that the schema has been upgraded (automatically) by Realm
-            if (oldSchemaVersion < kTalkDatabaseSchemaVersion) {
-                   // Nothing to do!
-                   // Realm will automatically detect new properties and removed properties
-                   // And will update the schema on disk automatically
-               }
         };
         
         // Tell Realm to use this new configuration object for the default Realm
--- a/NextcloudTalk/NCDatabaseManager.m
+++ b/NextcloudTalk/NCDatabaseManager.m
@@ -31,7 +31,7 @@
 
 NSString *const kTalkDatabaseFolder                 = @"Library/Application Support/Talk";
 NSString *const kTalkDatabaseFileName               = @"talk.realm";
-uint64_t const kTalkDatabaseSchemaVersion           = 26;
+uint64_t const kTalkDatabaseSchemaVersion           = 27;
 
 NSString * const kCapabilitySystemMessages          = @"system-messages";
 NSString * const kCapabilityNotificationLevels      = @"notification-levels";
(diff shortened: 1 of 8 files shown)


NextcloudTalk/RoomsTableViewController.m
```diff
@@ -1139,6 +1139,22 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
...
+                    [cell setUserStatusIcon:room.statusIcon];
+                }
+                else {
+                    [cell setUserStatus:room.status];
+                }
+            }else {
+                    [cell setUserStatus:room.status];
+                }
+            }
+        }
```
SystemKeeper: There's something wrong with the indentation here. Also the `else` should be in the same line as closing/opening bracets and have a space around it.
I would  make this a bit shorter, if I'm not mistaken, the following should do the same:

```objective-c
    //Show User Status
    if (room.type == kNCRoomTypeOneToOne && [room.status length] != 0) {
        if (![room.status isEqual:@"dnd"] && [room.statusIcon length] != 0) {
            [cell setUserStatusIcon:room.statusIcon];
        } else {
            [cell setUserStatus:room.status];
        }
    }
```

NextcloudTalk/RoomTableViewCell.h
```diff
@@ -36,10 +36,11 @@ extern CGFloat const kRoomTableCellHeight;
...
 @property (weak, nonatomic) IBOutlet UIImageView *favoriteImage;
 @property (weak, nonatomic) IBOutlet UIImageView *userStatusImageView;
+@property (weak, nonatomic) IBOutlet UILabel *userStatusLabel;
 
 @property (nonatomic, assign) BOOL titleOnly;
 
 - (void)setUnreadMessages:(NSInteger)number mentioned:(BOOL)mentioned groupMentioned:(BOOL)groupMentioned;
 - (void)setUserStatus:(NSString *)userStatus;
-
+-(void)setUserStatusIcon:(NSString *)userStatusIcon;
```
SystemKeeper: Space before `(void)` to be inline with the other part of the file.

NextcloudTalk/NCRoom.m
```diff
@@ -68,6 +68,9 @@ + (instancetype)roomWithDictionary:(NSDictionary *)roomDict
     room.hasCall = [[roomDict objectForKey:@"hasCall"] boolValue];
     room.canLeaveConversation = [[roomDict objectForKey:@"canLeaveConversation"] boolValue];
     room.canDeleteConversation = [[roomDict objectForKey:@"canDeleteConversation"] boolValue];
+    room.status = ([roomDict objectForKey:@"status"] != [NSNull null]) ? [[roomDict objectForKey:@"status"] stringValue] : @"";
```
SystemKeeper, line 71: I'm not absolutely sure about this one. Do we need to check for `[NSNull null]` here? I would imagine that
`[roomDict objectForKey:@"status"] ? [[roomDict objectForKey:@"status"] stringValue] : @""`

NextcloudTalk/NCDatabaseManager.m
```diff
@@ -89,6 +89,11 @@ - (id)init
         configuration.schemaVersion = kTalkDatabaseSchemaVersion;
         configuration.migrationBlock = ^(RLMMigration *migration, uint64_t oldSchemaVersion) {
             // At the very minimum we need to update the version with an empty block to indicate that the schema has been upgraded (automatically) by Realm
+            if (oldSchemaVersion < kTalkDatabaseSchemaVersion) {
```
SystemKeeper: Is this just to make it more explicit what's happening here?

SystemKeeper (changes requested):
Sorry for being nitpicky about the style 😄

NextcloudTalk/NCRoom.m
alazarevic5: I am not sure. I kept as it was, for now, because if we do not compare with null, when we receive null for status (for e.g. group conversation), app crashes, because we are trying to use null value.

Ivansss (approved):

PR #663, opened 2021-11-16
Warning for empty group room name

alazarevic5:
Present an alert when group room name is replaced with empty string.
Just first PR

### DIFF
--- a/NextcloudTalk/RoomInfoTableViewController.m
+++ b/NextcloudTalk/RoomInfoTableViewController.m
@@ -618,6 +618,15 @@ typedef enum FileAction {
     }
     if ([newRoomName isEqualToString:@""]) {
         _roomNameTextField.text = _room.name;
+        UIAlertController* alert = [UIAlertController alertControllerWithTitle:@"Warning"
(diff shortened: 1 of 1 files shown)


NextcloudTalk/RoomInfoTableViewController.m
```diff
@@ -618,6 +618,15 @@ - (void)renameRoom
     }
     if ([newRoomName isEqualToString:@""]) {
         _roomNameTextField.text = _room.name;
+        UIAlertController* alert = [UIAlertController alertControllerWithTitle:@"Warning"
+                                       message:@"Room name cannot be empty."
```
SystemKeeper: Localization is missing

SystemKeeper (changes requested):

Ivansss (approved):

NextcloudTalk/RoomInfoTableViewController.m
Ivansss: Changed strings and they are now localizable.

PR #671, opened 2021-11-19
Add deep link url parameter for room token

thisIsTheFoxe:
Allow other apps to open (and join) a specific, already existing room.
If both userId & roomToken are specified, prefer userId.

**Parameters**:
- _server_: The Nextcloud Server URL
- _user_: Your UserID, used for an account on _server_, that is logged into the Talk App
- _withUser_: The UserID of the user you want to talk to
- _withRoomToken_: The room token of an existing room on _server_

Usage:
```
nextcloudtalk://open-conversation?server={serverURL}&user={userId}&withUser={userId}

nextcloudtalk://open-conversation?server={serverURL}&user={userId}&withRoomToken={roomToken}
```

### Notes
If `withUser` or `withRoomToken` **is invalid**, opening the link basically just serves as a way to switch to another user.
If both `withUser` and `withRoomToken` **aren't present**, opening the link does nothing (no error, no account change).
Should someone be able to switch the users's account using `nextcloudtalk://open-conversation`? If so, `withUser` and `withRoomToken` should probably be optional. If not, the account should only be changed if the user or room has been verified. Wdyt?

## Related
- #656 

## TODO:
- [x] testing


NextcloudTalk/NCRoomsManager.m
```diff
@@ -804,8 +804,15 @@ - (void)joinOrCreateChatWithURL:(NSNotification *)notification
 {    
     NSString *accountId = [notification.userInfo objectForKey:@"accountId"];
     NSString *withUser = [notification.userInfo objectForKey:@"withUser"];
+    NSString *roomToken = [notification.userInfo objectForKey:@"withRoomToken"];
     [self checkForAccountChange:accountId];
-    [self joinOrCreateChatWithUser:withUser usingAccountId:accountId];
+
+    if (withUser != nil) {
+        [self joinOrCreateChatWithUser:withUser usingAccountId:accountId];
+    } else if (roomToken != nil) {
```
Ivansss: We could check `withUser` and `roomToken` as

```
if (withUser) {
    [self joinOrCreateChatWithUser:withUser usingAccountId:accountId];
} else if (roomToken) {
```
same as you did in NCUserInterfaceController.m ;)

Ivansss (changes requested):

Ivansss (approved):
Tested and works 👍

LexioJ:
IMG_1364
As this ability exists what now would be missing is to have a web landing page (as many other vendors provide) where the user is able to decide to join using their browser (use direct https) or dependent of the current connected platform (win/ios/mac/linux/android) to start the equivalent app protocol handler like nextcloudtalk://open-conversation to start the locally installed app instead. What do you think ? Where to raise this request ?

PR #672, opened 2021-11-19
Fix account check for custom urls.

Ivansss:
Try to get account from userId and server url instead of accountId.

### DIFF
--- a/NextcloudTalk/NCRoomsManager.m
+++ b/NextcloudTalk/NCRoomsManager.m
@@ -801,18 +801,11 @@ NSString * const NCRoomsManagerDidReceiveChatMessagesNotification   = @"ChatMess
 }
 
 - (void)joinOrCreateChatWithURL:(NSNotification *)notification
-{
-    NSURLComponents *urlComponents = [notification.userInfo objectForKey:@"url"];
-    NSArray *queryItems = urlComponents.queryItems;
-    NSString *server = [NCUtils valueForKey:@"server" fromQueryItems:queryItems];
-    NSString *user = [NCUtils valueForKey:@"user" fromQueryItems:queryItems];
-    NSString *withUser = [NCUtils valueForKey:@"withUser" fromQueryItems:queryItems];
-    NSString *accountId = [[NCDatabaseManager sharedInstance] accountIdForUser:user inServer:server];
-    TalkAccount *account = [[NCDatabaseManager sharedInstance] talkAccountForAccountId:accountId];
-    
+{    
+    NSString *accountId = [notification.userInfo objectForKey:@"accountId"];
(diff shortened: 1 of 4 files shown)


thisIsTheFoxe (changes requested):
lgtm so far, but this now fails (called via NotificationCenter): `[NCRoomsManager joinOrCreateChatWithURL]` because there it still makes the accountId from the `user` parameter in `talkAccountForAccountId`. 

This causes the app to request another login. Logging in causes *Account already added*. When canceling, the app now has a ghost user.
Furthermore, it is (still?) possible to login to the same account twice using 1.) the email address and 2.) the userId. Not sure if that's  intended and if it isn't if it's in the scope of this PR.. Your call.

Ivansss:
Fixed `joinOrCreateChatWithURL` in NCRoomsManager @thisIsTheFoxe 

> Furthermore, it is (still?) possible to login to the same account twice using 1.) the email address and 2.) the userId. Not sure if that's intended and if it isn't if it's in the scope of this PR.. Your call.

This will be fixed in another PR ;)

NextcloudTalk/NCRoomsManager.m
```diff
@@ -807,10 +807,9 @@ - (void)joinOrCreateChatWithURL:(NSNotification *)notification
     NSString *server = [NCUtils valueForKey:@"server" fromQueryItems:queryItems];
     NSString *user = [NCUtils valueForKey:@"user" fromQueryItems:queryItems];
     NSString *withUser = [NCUtils valueForKey:@"withUser" fromQueryItems:queryItems];
-    NSString *accountId = [[NCDatabaseManager sharedInstance] accountIdForUser:user inServer:server];
-    TalkAccount *account = [[NCDatabaseManager sharedInstance] talkAccountForAccountId:accountId];
+    TalkAccount *account = [[NCDatabaseManager sharedInstance] talkAccountForUserId:user inServer:server];
     
-    [self checkForAccountChange:accountId];
+    [self checkForAccountChange:account.accountId];
```
thisIsTheFoxe: *Technically* `account` can still return `nil`, right? (tho improbable, since it *should've been* checked before) 
Would it be possible to send the account directly via the `userInfo` of the notification?

thisIsTheFoxe (approved):
lgtm..
Sometimes it takes a second to load, but that shouldn't be a problem.

NextcloudTalk/NCRoomsManager.m
Ivansss: Modified it. Now `accountId` and `withUser` are sent in `userInfo`.
Let me know what do you think @thisIsTheFoxe ;)

thisIsTheFoxe: Looking good, thanks.. I also like the alert reacting 👍

PR #675, opened 2021-11-25
Swipe left to right to reply message

alazarevic5:
Swipe left to right in conversation to message reply.


Ivansss (approved):
Tested and works 👍

PR #681, opened 2021-12-02
Remember zoom-state per participant

### LINKED ISSUE #647: Keep zoom of CallParticipantViewCell after a user joins/leave
## Steps to reproduce
1. Have a call with multiple participants
2. Double tap each cell to set `_changeOriginalSize` to `true`
3. Let one participant leave the call

### Expected behaviour
All participants still in the call should keep their setting for `_changeOriginalSize`

### Actual behaviour
All `_changeOriginalSize` is reset to default value, because the cell is reused

SystemKeeper:
Fixes #647 

The zoom-state of a participant can be toggled by double-tapping on the corresponding CallParticipantViewCell. The state is remembered _per cell_, which means it gets set back to default whenever the cell is reused (for example when a participant joins or leaves the call). We now store the zoom-state in the NCPeerConnection and can therefore restore it, when a cell is reused.

Not sure if we should dispatch the following to the main-thread?

https://github.com/nextcloud/talk-ios/blob/8e904206264bbdbeb3fb2e9b0ce1922f2d6c27a8/NextcloudTalk/CallViewController.m#L1186-L1193

### DIFF
--- a/NextcloudTalk/CallParticipantViewCell.h
+++ b/NextcloudTalk/CallParticipantViewCell.h
@@ -30,6 +30,7 @@ extern CGFloat const kCallParticipantCellMinHeight;
 @class CallParticipantViewCell;
 @protocol CallParticipantViewCellDelegate <NSObject>
 - (void)cellWantsToPresentScreenSharing:(CallParticipantViewCell *)participantCell;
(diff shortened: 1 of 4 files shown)


Ivansss (approved):
Tested and works 👍 

Thanks for the PR @SystemKeeper ;)

PR #682, opened 2021-12-02
Show warning when server is in maintenance mode

### LINKED ISSUE #576: No feedback when the server is in maintenance mode
## Steps to reproduce
1. Have a nextcloud account
2. Put the instance in maintainence mode or update
3. While that is in maintenance mode open the Talk app

### Expected behaviour
A note that the server is in maintenance mode

### Actual behaviour
You just see the old/cached conversation list and are not aware why nothing is updating

SystemKeeper:
Fixes #576


Ivansss (approved):
Awesome 🎉

PR #684, opened 2021-12-03
Add all peers as participants, even if they don't publish streams

### LINKED ISSUE #401: Participants without camera/microphone are invisible
The following is tested with 10.0.1 and HPB backend.

* Start a 1:1 call from mobile (doesn't really matter which way around)
* On Talk Web join the call with disabled camera and microphone

**Expected**
The app should recognize the participant as joined

**Current**
The app does not recognize the participant and sits in the "Waiting for X to join call...", BUT already transmits voice/video. For the mobile user it looks like nobody is connected.

**Mobile**
image

**Web**
image

`CallState` is only changed to `CallStateInCall` when `_peersInCall` is not empty:
https://github.com/nextcloud/talk-ios/blob/858b6df8dbea6e875d1befcc610640befcf3518e/NextcloudTalk/CallViewController.m#L367-L378

This only changes when remote stream was added to the room:
https://github.com/nextcloud/talk-ios/blob/858b6df8dbea6e875d1befcc610640befcf3518e/NextcloudTalk/CallViewController.m#L936-L939

But without a stream -> no peer -> no participant.

I guess one way would be to expose/watch the number of `pendingUsersInRoom` from `NCCallController` in case there're no peers to determine the current `callState`. A `CallStateInCallWithoutStream` would then be a possibility. Not sure about internal signaling though.

SystemKeeper:
Attempt to fix #401 

When there are participants without streams, they are invisible in talk-ios at the moment. Now we add the peer immediately, without waiting for streams. Also moved setting properties of `CallParticipantViewCell` to a single function.
This should only affect external signaling as of now. 

I think the same should work for internal signaling:

https://github.com/nextcloud/talk-ios/blob/47bf3ebba51733afa18422f0949e8b914655cb0d/NextcloudTalk/NCCallController.m#L697-L704

We could call `getPeerConnectionWrapperForSessionId` in all cases. Any insights about this @Ivansss ?

### DIFF
--- a/NextcloudTalk/CallViewController.m
+++ b/NextcloudTalk/CallViewController.m
@@ -1200,6 +1200,25 @@ typedef NS_ENUM(NSInteger, CallState) {
     return [_peersInCall count];
 }
 
+- (void)updateParticipantCell:(CallParticipantViewCell *)cell withPeerConnection:(NCPeerConnection *)peerConnection
+{
+    BOOL isVideoDisabled = peerConnection.isRemoteVideoDisabled;
+    
+    if (_isAudioOnly || peerConnection.remoteStream == nil) {
+        isVideoDisabled = YES;
+    }
+    
+    [cell setVideoView:[_videoRenderersDict objectForKey:peerConnection.peerId]];
+    [cell setUserAvatar:[_callController getUserIdFromSessionId:peerConnection.peerId]];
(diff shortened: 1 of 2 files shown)


Ivansss (approved):
Perfect 👍 Thanks for the PR :)

SystemKeeper:
@Ivansss Any thoughts about doing the same with internal signaling?

PR #686, opened 2021-12-03
Try to get displayName as early as possible

SystemKeeper:
Try to get the display name of the participant as soon as we init `NCPeerConnection`.

This does not work for guests (at least with external signaling), because the displayName property is not available from the participant received by the signaling server.

### DIFF
--- a/NextcloudTalk/NCCallController.m
+++ b/NextcloudTalk/NCCallController.m
@@ -461,13 +461,6 @@ static NSString * const kNCVideoTrackKind = @"video";
         
         if (displayName) {
             [peerConnectionWrapper setPeerName:displayName];
-        } else {
-            // Fallback to userId, when displayName can't be resolved
-            NSString *userId = [self getUserIdFromSessionId:sessionId];
-            
-            if (userId) {
-                [peerConnectionWrapper setPeerName:userId];
-            }
         }
         
         if (![_externalSignalingController hasMCU] || !screensharingPeer) {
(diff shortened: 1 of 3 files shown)


NextcloudTalk/NCCallController.m
```diff
@@ -455,7 +455,21 @@ - (NCPeerConnection *)getPeerConnectionWrapperForSessionId:(NSString *)sessionId
...
+        
+        if (displayName) {
+            [peerConnectionWrapper setPeerName:displayName];
+        } else {
+            // Fallback to userId, when displayName can't be resolved
+            NSString *userId = [self getUserIdFromSessionId:sessionId];
+            
+            if (userId) {
+                [peerConnectionWrapper setPeerName:userId];
+            }
```
Ivansss: Not sure about this. 
`userId` might be meaningful in some NC instances, but (e.g. when using LDAP or SAML) the `userId` could be a string of random characters or numbers.

Ivansss (changes requested):

NextcloudTalk/NCCallController.m
SystemKeeper: I see.. so just remove the fallback?

Ivansss: I guess...what do you think? :)

SystemKeeper: Yep, totally agree 👍 Haven't thought about LDAP or SAML in this case.

Ivansss (approved):
Nice adding! Thanks @SystemKeeper 👍

Issue #667, opened 2021-11-18
Show conversation name in call view for group conversations


Ivansss:
Display the name of the conversation name in group calls.

SystemKeeper:
Fixed with #1038

PR #687, opened 2021-12-03
Call view design enhancements

alazarevic5:
- [x] Show conversation name in group calls.
- [x] Set participant name at the bottom-left corner.
- [x] Set mute and screensharing indicators at the bottom right corner.
- [x] Add rounded corners to cells (same as in the web UI)
- [ ] Call control buttons should push and shrink the collection view.


SystemKeeper:
A few things I noticed:

* CustomHeaderView.* files are not added to project / changes to project are not committed (this might be on purpose, though, just checking you left that out intentionally and want to add it later?)

* Label does not get wrapped anymore:
image

* Background-color mismatch while waiting (does not look intended?):
image

SystemKeeper:
Fixes #667 I guess?

alazarevic5:
I didn't push correctly. I'll have to correct it and send a new commit.

NextcloudTalk/CallParticipantViewCell.xib
```diff
@@ -25,7 +25,7 @@
                         <autoresizingMask key="autoresizingMask" flexibleMinX="YES" flexibleMaxX="YES" flexibleMinY="YES" flexibleMaxY="YES"/>
                     </imageView>
                     <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" fixedFrame="YES" text="" textAlignment="center" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="D5h-T2-aB9">
-                        <rect key="frame" x="8" y="32" width="184" height="30"/>
+                        <rect key="frame" x="7" y="32" width="185" height="30"/>
```
Ivansss: Is this a needed change?

NextcloudTalk/CallViewController.m
```diff
@@ -190,6 +193,11 @@ - (void)viewDidLoad
     
     [self.collectionView registerNib:[UINib nibWithNibName:kCallParticipantCellNibName bundle:nil] forCellWithReuseIdentifier:kCallParticipantCellIdentifier];
     
+    [self.collectionView registerClass:CustomHeaderView.self forSupplementaryViewOfKind:UICollectionElementKindSectionHeader withReuseIdentifier:@"HeaderCollectionReusableView"];
+    
+    UICollectionViewFlowLayout *layout = self.collectionView.collectionViewLayout;
```
Ivansss: This shows a warning
Could be cast as:
UICollectionViewFlowLayout *flow = (UICollectionViewFlowLayout *) self.collectionView.collectionViewLayout;

NextcloudTalk/CallViewController.m
```diff
@@ -190,6 +193,11 @@ - (void)viewDidLoad
...
+    layout.sectionHeadersPinToVisibleBounds = YES;
```
Ivansss: Doesn't seem to make the header sticky :S

NextcloudTalk/CallViewController.m
```diff
@@ -1233,6 +1241,28 @@ -(void)collectionView:(UICollectionView *)collectionView willDisplayCell:(UIColl
...
+    if (kind == UICollectionElementKindSectionHeader) {
+        CustomHeaderView *headerView = [collectionView dequeueReusableSupplementaryViewOfKind:UICollectionElementKindSectionHeader withReuseIdentifier:@"HeaderCollectionReusableView" forIndexPath:indexPath];
+        [headerView configure];
+        headerView.backgroundColor = UIColor.blackColor;
+        [headerView setConversationNameWithName:_room.displayName];
+        reusableview = headerView;
+    }
+    return reusableview;
+}
+
```
Ivansss: Not sure about placing the conversation name view as collectionView's section header.
I was thinking about a floating view on top of the collection view.
But it could work (to leave it as collectionView header) if we make it sticky and we place the toggle chat button in it.
Wat do you think @alazarevic5 ? :)

Ivansss (changes requested):

SystemKeeper:
Last commit fixed project file and label-wrapping. Still not sure about the look of it (my 2nd screenshot). Should the safe are above be black like in aleksandras screenshot?

NextcloudTalk/CallViewController.m
alazarevic5: Yes I have removed header. Instead of that I put label for title above collection view for now.

NextcloudTalk/CallViewController.m
```diff
@@ -1210,7 +1217,11 @@ - (void)updateParticipantCell:(CallParticipantViewCell *)cell withPeerConnection
     
     [cell setVideoView:[_videoRenderersDict objectForKey:peerConnection.peerId]];
     [cell setUserAvatar:[_callController getUserIdFromSessionId:peerConnection.peerId]];
-    [cell setDisplayName:peerConnection.peerName];
+    if (_room.type != kNCRoomTypeOneToOne) {
```
SystemKeeper, line 1234: Do we need to do the same in `didReceiveNick` ?

alazarevic5:
For now looks like this.. I need to find way for reposition of mic icon and participant name in cell.
Simulator Screen Shot - iPhone 12 Pro Max - 2021-12-08 at 21 01 59

SystemKeeper:
Maybe move it to the bottom left of the participantCell?

alazarevic5:
I have placed bottom left, but it floating menu with controls sometimes is a bit above name
Simulator Screen Shot - iPhone 12 Pro Max - 2021-12-09 at 21 57 09

SystemKeeper:
Yes that’s way easier in web, where the  controls are on a top bar. When we do the same sometime in the future, this problem vanishes.

For reference, here is a PR from web https://github.com/nextcloud/spreed/pull/6648

I think you can save some height with moving the muted icon behind the participant name (see the web PR above). Maybe @Ivansss  has some input on this?

Ivansss:
> For reference, here is a PR from web nextcloud/spreed#6648
> 
> I think you can save some height with moving the muted icon behind the participant name (see the web PR above).

I also think the name and the mute indicator should be at the same position as in the web app.

My only concern is for the cells that are at the bottom like "Aleksadra 2" in the screenshot above.
It might be that call buttons stay on top of the participant name in some situations.

@nextcloud/designers-talk any idea here?

SystemKeeper:
> It might be that call buttons stay on top of the participant name in some situations.

Maybe have it bottom left per default and in the bottom row, move it above the controls ? Like the camera view which gets lifted..

alazarevic5:
I will put mute indicator in the left side in cell like on the web. Yes, as @Ivansss said, sometimes are buttons in front of the name. I can move name a bit higher but in that case is not anymore bottom left corner.

Ivansss:
> I will put mute indicator in the left side in cell like on the web.

I think the participant name should be on the left side and the indicators on the right.

|  participant name | mute indicator | screensharing indicator |
| --- | --- | --- |

If there are no indicators the participant name label fills the whole bottom area of the cell.
We could implement that behaviour with a horizontal stack view.

> Maybe have it bottom left per default and in the bottom row, move it above the controls ? Like the camera view which gets lifted..

Could be an option :)

SystemKeeper:
While we're at it, should we remove the background of the indicators as well? (Like in web?)

alazarevic5:
Now is name on the left and indicators on the right in a stack view. Because of stack view I have commented two lines in project for reposition of icons, because they are now in a stack. Only problem are maybe bottom buttons.
Screenshot 2021-12-11 at 19 20 26

alazarevic5:
Added rounded corners
Collection view is on the whole view from side to side.
Simulator Screen Shot - iPhone 12 Pro Max - 2021-12-14 at 18 40 22

alazarevic5:
Call control buttons now shrink CollectionView. 
In audio calls buttons are on the bottom fixed, but in video call collection view is on whole window and buttons are showing and disappearing on tap or after a few seconds.
Cells of participants now have rounded corners and name and mic icon are on the bottom of the cell.

jancborchardt:
Since there’s now the rounded corners layout as in the web, at least for audio calls where there’s never video we can also leave space for the top part (title) and the bottom part (controls), right? :) So don’t extend the tiles all the way down and up to the end of the screen in this case.

For video calls, it might still be nice to use the whole space, not fully sure about that. cc @marcoambrosini

marcoambrosini:
Yes @jancborchardt, totally agree. @Ivansss and I had a call about it and all that will come at some point. I'm just not sure wether it will happen in this pr or not

alazarevic5:
> 

Yes, it will be changed like your idea, to push and shrink collection view with buttons. Because now they are fixed with two cases.

Issue #664, opened 2021-11-17
Typing and "sending" without server-connection


lotofbits:
Hey there,

currently it is only possible to type messages when the app has connected to the server. Sometimes it takes a few ms to connect, which doesn't feel smooth, other times there may not be an internet connection at all, but you want to get a thought out.

In my opinion, it would greatly improve the user experience to be able to type the message at any time.

It should also be possible to "send" the message without a server connection. All that is needed is a hint that there is currently no connection, but that the message will be sent as soon as there is a connection again. It should therefore also be possible to send the messages in the background.

PR #688, opened 2021-12-03
Leave offline mode when network availability changes

### LINKED ISSUE #533: Offline -> online: no chat possible
- go into chat
- type something
- swipe up and enable airplane mode
- see that app is offline ✅
- swipe up to disable airplane mode
- see that app is online
- but cannot send a message, as input field is still disabled

SystemKeeper:
Fixes #533

I don't think the other way round is currently needed, therefore I just check for `kConnectionStateConnected`.


SystemKeeper:
#664 is related, we could look into this in a follow-up PR.

Ivansss (approved):
Tested and works 👍 
Thanks for this nice adding @SystemKeeper :)

PR #690, opened 2021-12-08
Call view enhancements

Ivansss:
- [x] Do not reload the whole collection view when adding, removing or modifying a peer.


alazarevic5 (approved):
It worked fine. I made modifications as in description and worked as expected.

Issue #693, opened 2021-12-15
No Call notifications when app is closed


DwinAnderson:
## Steps to reproduce
1. Two iPhones XR, Nextcloud Talk is opened on both devices.
2. I minimize the application and close it on both devices.
3. I choose one of two phones (randomly, as an example let it be "phone №2"), open the application through the application icon, then go to the chat with user and make a 1 to 1 call.
4. No call notification on "phone №1". 
Remarks:
1. When app is opened on both phones, there is no such problem! 
2. Message notifications arrive in both cases (when users are in app at this moment and even they are not, and even apps are not launched in the background at all).

### Expected behaviour
Call notification should be even app is not launched or even it is launched but in the background.

### Actual behaviour
No call notification when app launched in the background or when it is not launched.

## Device information

**Device:** iPhone XR number 1 and iPhone XR number 2

**iOS version:** both have iOS 15.1.

**Talk version:** both have 13.0.1.

## Server information

**Nextcloud version:** 22.2.3

**Talk version:** 12.2.2

**Custom Signaling server configured:** no

**Custom TURN server configured:** no

**Custom STUN server configured:** no

DwinAnderson:
All the previous issues about problems with call notifications were closed or forwarded to chat notifications problems. This problem wasn’t solved for long times. Please, is it possible at all to get normal call notifications? Maybe settings for personal notification server?
Btw Android users have fix for this notifications problem from 2018th already, then they had problems again in 2020th but everything was solved again. 
https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server
” You’re responsible for setting up a provider server (or servers) and for configuring your app to handle notifications on the user’s device. Apple manages everything in between, including the presentation of notifications to the user.”
May be notifications are not safe at all?
I’ve read about sideloading apps, is it possible for nextcloud talk, so I could configure it in my way to get all the notifications from my personal apn server?

SystemKeeper:
Hey DwinAnderson,

as far as I can see, there're no known issues regarding push notifications. I just tested the scenario you described and it worked as usual.
You might want to check the following document about `occ notification:test-push` and see what happens: https://github.com/nextcloud/talk-android/blob/master/docs/notifications.md#-check-server-settings
Other than that try to remove your account from talk-ios and re-add it. Also check your server logs for any errors or exceptions. 
If you can access the console of your iPhone (via MacOS), you might want to check for any messages from `NextcloudTalk` or anything regarding push.

[...]

PR #694, opened 2021-12-17
Correctly process test notifications

SystemKeeper:
Sending a test-notification through the following command: `occ notification:test-push --talk <userid>` results in "You received a new notification that could not be decrypted". This PR correctly processes those notifications. It makes also sure, that we don't wait for the extension timeout when we already know that nothing is happening anymore.

Related #693 

For reference, this is the payload which is send via the notification app:
```json
{
   "nid":32530,
   "app":"admin_notification_talk",
   "subject":"Testing push notifications",
   "type":"admin_notifications",
   "id":"61bc80a9"
}
```

### DIFF
--- a/NextcloudTalk/NCPushNotification.h
+++ b/NextcloudTalk/NCPushNotification.h
@@ -28,7 +28,8 @@ typedef NS_ENUM(NSInteger, NCPushNotificationType) {
     NCPushNotificationTypeRoom,
     NCPushNotificationTypeChat,
     NCPushNotificationTypeDelete,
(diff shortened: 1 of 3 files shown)


DwinAnderson:
May I be helpful somehow in here?

SystemKeeper:
Thanks @DwinAnderson ! This is quite tricky to test before it's released, because it needs to be signed by the nextcloud org to be able to receive push notifications. What would be great, if this gets approved and merged to get your feedback after the release to the appstore.

Ivansss (approved):

PR #695, opened 2021-12-17
Set the default realm-configuration when used in ShareExtension

SystemKeeper:
This fixes the ShareExtension crashing with a "migration required" exception, although the log reports the same schema version.

This problem only occurs when multiple accounts are configured in talk-ios and a new version is installed with a higher schema version. The main problem comes from this line:
https://github.com/nextcloud/talk-ios/blob/901b57cd6237483817593532152bd328a2e2391d/ShareExtension/ShareViewController.m#L235
When multiple accounts are set up, NCAPIController is used to retrieve the profile image of the account. NCAPIController (when initalizing) iterates through the TalkAccounts, but with default configuration of realm (therefore creating a realm database in the ShareExtension folder and not using the one in the AppGroup). As realm isn't opened explicitly, no migration is happening on the database and the ShareExtension never loads. 

After removing accounts and making sure that there's only one account left, ShareExtension starts working again (basically confirming the above).

Stacktraces leading to the exception:
image
image

Things to consider:
We might want to think about this explicit realm usage in ShareViewController (`_realm`), because (as far as I can see) this shouldn't be neccessary in this case anymore.

### DIFF
--- a/ShareExtension/ShareViewController.m
+++ b/ShareExtension/ShareViewController.m
@@ -109,7 +109,11 @@
         // At the very minimum we need to update the version with an empty block to indicate that the schema has been upgraded (automatically) by Realm
     };
     NSError *error = nil;
+    
+    // Set the default configuration to make sure we always use the correct realm-file
+    [RLMRealmConfiguration setDefaultConfiguration:configuration];
     _realm = [RLMRealm realmWithConfiguration:configuration error:&error];
+    
     [self setupShareViewForAccount:nil];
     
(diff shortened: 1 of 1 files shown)


SystemKeeper:
We might also consider doing the same in NotificationServiceExtension. There's no such case as in ShareExtension just yet, but I can imagine this can be easily missed as seen in ShareExtension.

Ivansss (approved):
Tested and works 👍

PR #700, opened 2022-01-07
Handle error while requesting call transaction.

Ivansss:
Starting a call from the chat view using the top right buttons fails some times (usually only on the first try after launching the app)
`The operation couldn’t be completed. (com.apple.CallKit.error.requesttransaction error 7.)`

This causes the call (or videocall) button keeps spinning forever.
This PR resets the call buttons if requesting the call transaction fails.

### DIFF
--- a/NextcloudTalk/CallKitManager.h
+++ b/NextcloudTalk/CallKitManager.h
@@ -28,6 +28,7 @@ extern NSString * const CallKitManagerDidEndCallNotification;
 extern NSString * const CallKitManagerDidStartCallNotification;
 extern NSString * const CallKitManagerDidChangeAudioMuteNotification;
 extern NSString * const CallKitManagerWantsToUpgradeToVideoCall;
+extern NSString * const CallKitManagerDidFailRequestingCallTransaction;
 
 @interface CallKitCall : NSObject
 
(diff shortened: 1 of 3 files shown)


SystemKeeper (changes requested):
I receive the following exception when error 7 occurs:

image

image

SystemKeeper:
As discussed, looks like this is a timing issue between first initializing CallKit vs. requesting a transaction. In a quick test, the following was enough to make the error 7 disappear:

```obj-c
        dispatch_async(dispatch_get_main_queue(), ^{
            [self.callController requestTransaction:transaction completion:^(NSError * _Nullable error) {
                if (!error) {
                    [weakSelf.calls setObject:call forKey:callUUID];
                } else {
                    NSLog(@"%@", error.localizedDescription);
                    NSDictionary *userInfo = [NSDictionary dictionaryWithObject:token forKey:@"roomToken"];
                    [[NSNotificationCenter defaultCenter] postNotificationName:CallKitManagerDidFailRequestingCallTransaction
                                                                        object:self
                                                                      userInfo:userInfo];
                }
            }];
        });
```

Also tried
`dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0.5 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^(void)`, which worked as well.

A retry should be added nonetheless in another pr.

Ivansss:
I just dispatch the layout modification code on the main thread for now.
Still can't trigger the error 7 reliably 😅 but as we discussed we can dispatch the requestTransaction with a small delay and retry if it fails in another PR ;)

SystemKeeper (approved):

PR #704, opened 2022-01-12
Moving SettingsViewController to Swift

alazarevic5:


SystemKeeper:
Hey @alazarevic5 👋 
Please update your license header to the GNU GPL one, as in other files

alazarevic5:
@SystemKeeper Thanks, I have updated.

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,676 @@
...
+                self.tableView.reloadData()
+            }
+        }
+    }
+
+    // MARK: Notifications
+
+    @objc func appStateHasChanged(notification: NSNotification) {
+        let appState = notification.userInfo!["appState"]
+        self.adaptInterfaceForAppState(appState: appState as? AppState ?? kAppStateReady)
```
Ivansss: only adapt UI if we receive a valid appState

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,676 @@
...
+            break
+        }
+    }
+
+    // MARK: Profile actions
+
+    func userProfilePressed() {
+        let activeAccount = NCDatabaseManager.sharedInstance().activeAccount()
+        let userProfileVC = UserProfileViewController(account: activeAccount)
+        self.navigationController!.pushViewController(userProfileVC, animated: true)
```
Ivansss: Do not use force unwrapping

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,676 @@
...
+    }
+
+    // MARK: User Status
+
+    func presetnUserStatusOptions() {
+        let viewController = UserStatusTableViewController(userStatus: activeUserStatus!)
+        self.navigationController!.pushViewController(viewController, animated: true)
```
Ivansss: Do not use force unwrapping

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,676 @@
...
+    }
+
+    func presentSetPhoneNumberDialog() {
+        let alertTitle = NSLocalizedString("Phone number", comment: "")
+        let alertMessage = NSLocalizedString("You can set your phone number so other users will be able to find you", comment: "")
+        let setPhoneNumberDialog = UIAlertController(title: alertTitle, message: alertMessage, preferredStyle: .alert)
+
+        setPhoneNumberDialog.addTextField { [self] textField in
+            let location = NSLocale.current.regionCode
+            textField.text = "+\(self.phoneUtil?.getCountryCode(forRegion: location) ?? 0)"
```
Ivansss: Only set a country code if we could find one.
Setting "+0" would be confusing for the user.

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,676 @@
...
+            do {
+                let exampleNumber: NBPhoneNumber =  try phoneUtil!.getExampleNumber(location)
+                textField.placeholder = try phoneUtil?.format(exampleNumber, numberFormat: NBEPhoneNumberFormat.INTERNATIONAL)
+            } catch {
+                print("Error with phone number format...")
```
Ivansss: Do we really need a try-catch block here?
I would just set the example number as placeholder if we are able to get one.

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,676 @@
...
+        setPhoneAction = UIAlertAction(title: NSLocalizedString("Set", comment: ""), style: .default, handler: { _ in
+            let phoneNumber = setPhoneNumberDialog.textFields?[0].text
+
+            NCAPIController.sharedInstance().setUserProfileField(kUserProfilePhone, withValue: phoneNumber, for: NCDatabaseManager.sharedInstance().activeAccount()) { error, _ in
+                if error != nil {
+                    self.presentPhoneNumberErrorDialog(phoneNumber: phoneNumber!)
+                    print("Error setting phone number ", error ?? "")
+                } else {
+                    let alert = UIAlertController(title: NSLocalizedString("Phone number set successfully", comment: ""), message: "", preferredStyle: .alert)
+                    alert.addAction(UIAlertAction(title: "OK", style: .default, handler: nil))
```
Ivansss: This could be a toast like in the objective C view controller

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,676 @@
...
+    }
+
+    func presentPhoneNumberErrorDialog(phoneNumber: String) {
+        var failedPhoneNumberDialog: UIAlertController?
+        var failedPhoneNumber: NBPhoneNumber?
+        do {
+            failedPhoneNumber = try phoneUtil!.parse(phoneNumber, defaultRegion: nil)
+        } catch {
+            print("Error...")
+        }
```
Ivansss: Do we really need a try-catch block here?
If yes, please add a more descriptive error message.

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,676 @@
...
+                   title: alertTitle,
+                   message: alertMessage + "\(try phoneUtil!.format(failedPhoneNumber, numberFormat: NBEPhoneNumberFormat.INTERNATIONAL))",
+                   preferredStyle: .alert)
+            let retryAction = UIAlertAction(title: NSLocalizedString("Retry", comment: ""), style: .default, handler: { [self] _ in presentSetPhoneNumberDialog()
+            })
+            failedPhoneNumberDialog!.addAction(retryAction)
+
+        } catch {
+            print("Error...")
+        }
```
Ivansss: Do we really need a try-catch block here?
If yes, please add a more descriptive error message.

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,676 @@
...
+    // MARK: UITextField delegate
+
+    func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
+        if textField.tag == kPhoneTextFieldTag {
+            let inputPhoneNumber = (textField.text as NSString?)?.replacingCharacters(in: range, with: string)
+            var phoneNumber: NBPhoneNumber?
+            do {
+                phoneNumber = try phoneUtil!.parse(inputPhoneNumber, defaultRegion: nil)
+            } catch {
+            }
```
Ivansss: Do we really need a try-catch block here?

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,676 @@
...
+            return 100
+        }
+        return 48
+    }
+
+    override func tableView(_ tableView: UITableView, titleForHeaderInSection section: Int) -> String? {
+        let sections = getSettingsSections()
+        let settignsSection = sections[section]
+
+        switch settignsSection {
```
Ivansss: correct typo settignsSection -> settingsSection

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,676 @@
...
+        default:
+            break
+        }
+        return nil
+    }
+
+    override func tableView(_ tableView: UITableView, titleForFooterInSection section: Int) -> String? {
+        let sections = getSettingsSections()
+        let settignsSection = sections[section]
+        if settignsSection == SettingsSection.kSettingsSectionAbout.rawValue {
```
Ivansss: correct typo settignsSection -> settingsSection

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,676 @@
...
+            let option = options[indexPath.row]
+            switch option {
+            case ConfigurationSectionOption.kConfigurationSectionOptionVideo.rawValue:
+                self.presentVideoResoultionsSelector()
+            case ConfigurationSectionOption.kConfigurationSectionOptionBrowser.rawValue:
+                self.presentBrowserSelector()
+            case ConfigurationSectionOption.kConfigurationSectionOptionReadStatus.rawValue:
+                break
+            case ConfigurationSectionOption.kConfigurationSectionOptionContactsSync.rawValue:
+                break
```
Ivansss: not needed if we have the default case

Ivansss (changes requested):

NextcloudTalk/SettingsTableViewController.swift
alazarevic5: Ok have added now if let syntax with check

NextcloudTalk/SettingsTableViewController.swift
alazarevic5: self.view.makeToast(NSLocalizedString("Phone number set successfully", comment: ""), duration: 3, position: CSToastPositionCenter)

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,663 @@
...
+
+    override func tableView(_ tableView: UITableView, titleForFooterInSection section: Int) -> String? {
+        let sections = getSettingsSections()
+        let settingsSection = sections[section]
+        if settingsSection == SettingsSection.kSettingsSectionAbout.rawValue {
+            let appName = (Bundle.main.object(forInfoDictionaryKey: "CFBundleDisplayName") as? String)!
+            let appVersion = (Bundle.main.object(forInfoDictionaryKey: "CFBundleShortVersionString") as? String)!
+            return appName + appVersion + copyright
+        }
+        return nil
```
Ivansss: Missing some footers. Check the objective C settings implementation.

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,663 @@
...
+        self.navigationController?.pushViewController(viewController, animated: true)
+    }
+
+    // MARK: User phone number
+
+    func checkUserPhoneNumber() {
+        NCSettingsController.sharedInstance().getUserProfile { _ in
+            let activeAccount = NCDatabaseManager.sharedInstance().activeAccount()
+            if activeAccount.phone.isEmpty {
+                self.presentSetPhoneNumberDialog()
```
Ivansss, line 239: Was not triggered the first time I changed contactSync value to enabled. Please check.

Ivansss (changes requested):

NextcloudTalk/SettingsTableViewController.swift
alazarevic5: I made a change

NextcloudTalk/SettingsTableViewController.swift
alazarevic5: footers added.

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,699 @@
...
+            return appName + appVersion + copyright
+        }
+        if let activeUserStatus = activeUserStatus {
+            if settingsSection == SettingsSection.kSettingsSectionUserStatus.rawValue && activeUserStatus.status == kUserStatusDND {
+                return NSLocalizedString("All notifications are muted", comment: "")
+            }
+        }
+        if let contactSyncSwitch = contactSyncSwitch {
+            if settingsSection == SettingsSection.kSettingsSectionConfiguration.rawValue && contactSyncSwitch.isOn {
+                if NCContactsManager.sharedInstance().isContactAccessDetermined() && NCContactsManager.sharedInstance().isContactAccessAuthorized() {
```
Ivansss: The check should be
`if NCContactsManager.sharedInstance().isContactAccessDetermined() && !NCContactsManager.sharedInstance().isContactAccessAuthorized()`

Ivansss (changes requested):

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,699 @@
...
+            cell.serverAddressLabel.text = accountServer.replacingOccurrences(of: "https://", with: "")
+            cell.userImageView.image = NCAPIController.sharedInstance().userProfileImage(for: activeAccount, with: CGSize(width: 160, height: 160))
+            cell.accessoryType = .disclosureIndicator
+            return cell
+        case SettingsSection.kSettingsSectionUserStatus.rawValue:
+            cell = UITableViewCell(style: .default, reuseIdentifier: userStatusCellIdentifier)
+            if activeUserStatus != nil {
+                cell.textLabel?.text = activeUserStatus!.readableUserStatus()
+                let statusImage = activeUserStatus!.userStatusImageName(ofSize: 24)
+                cell.imageView?.image = UIImage(named: statusImage)
```
Ivansss: This part is not showing the user status message (if there is any) as in the objective C view controller.

Ivansss (changes requested):

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,705 @@
...
+
+    func userProfilePressed() {
+        let activeAccount = NCDatabaseManager.sharedInstance().activeAccount()
+        let userProfileVC = UserProfileViewController(account: activeAccount)
+        self.navigationController?.pushViewController(userProfileVC, animated: true)
+    }
+
+    // MARK: User Status
+
+    func presetnUserStatusOptions() {
```
nickvergessen: typo?

alazarevic5: Ok I will correct character mistake, thanks

.swiftlint.yml
```diff
@@ -16,6 +16,10 @@ line_length:
 type_body_length:
   # error: 350
   error: 500
+  
+function_body_length:
+  error: 200
```
thisIsTheFoxe: Personally, I find that functions with more than 100 lines of code become increasingly complex and difficult to read. And it's also harder to write encapsulated unit tests for such functions. Just imo, tho..

NextcloudTalk/SettingsUserTableViewCell.swift
```diff
@@ -0,0 +1,49 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+import UIKit
+import NCCommunication
+
+let kUserSettingsCellIdentifier = "UserSettingsCellIdentifier"
+let kUserSettingsTableCellNibName = "UserSettingsTableViewCell"
```
thisIsTheFoxe: I like to keep constants like those directly attached to the type instead of having them being global variables.
I.e.: [Example](https://tio.run/##dY6xCsJAEAX7/YrlKm2srAQRsbJJY7S/xI0ubi5yuxhE/PYzV4gasJ4Z3tOeG5unVItXxR2ZcTjpXimWvhI6MPUbEsEHIKp54xqFDC/bIwXjhiniEl32323WP9SNukwLrgrf0jj8GXQwhDcfUbuWyvPAF7gO9xU8Aa6Rg03@fp19nZum9AI)

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,705 @@
...
+    case kConfigurationSectionOptionContactsSync
+}
+
+enum AboutSection: Int {
+    case kAboutSectionPrivacy = 0
+    case kAboutSectionSourceCode
+    case kAboutSectionNumber
+}
+
+let kPhoneTextFieldTag = 99
```
thisIsTheFoxe: Global variables are definitely better better than magic numbers.
But I'm not sure if this one needs to be global? It can just live inside the `SettingsTableViewController` can't it?

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,705 @@
...
+        if NCDatabaseManager.sharedInstance().serverHasTalkCapability(kCapabilityPhonebookSearch) {
+            options.append(ConfigurationSectionOption.kConfigurationSectionOptionContactsSync.rawValue)
+        }
+        return options
+    }
+
+    func getSectionForSettingsSection(section: SettingsSection) -> Int {
+        let sectionNumber = getSettingsSections().firstIndex(of: section.rawValue)
+        if NSNotFound != sectionNumber {
+            return sectionNumber!
```
thisIsTheFoxe: I'm not exactly sure what this does since `NSNotFound == Int.max` and if the index can't be found it's `nil` so the force unwrap will fail (because *If an optional has a value, it’s considered to be “not equal to” `nil`*.
I guess you want to make sure the index was found or else return 0?
Easiest would be to use a default value:
```swift
  return sectionNumber ?? 0
```
Or, if you need section number to do some other work:
```swift
if let sectionNumber = sectionNumber {
  return (sectionNumber + 0) * 1
}
```

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,705 @@
...
+            self.tableView.reloadData()
+        }
+        self.getActiveUserStatus()
+    }
+
+    func getActiveUserStatus() {
+        let activeAccount = NCDatabaseManager.sharedInstance().activeAccount()
+        NCAPIController.sharedInstance().getUserStatus(for: activeAccount) { userStatus, error in
+            if error == nil && userStatus != nil {
+                self.activeUserStatus = NCUserStatus(dictionary: userStatus!)
```
thisIsTheFoxe: Use safe unwrapping wherever possible. Btw, I actually like to use commas instead of && in general because I think it makes the order of operation easier to follow (`||` before `,`)

```suggestion
            if let userStatus = userStatus, error == nil {
                self.activeUserStatus = NCUserStatus(dictionary: userStatus)
```

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,705 @@
...
+                self.tableView.reloadData()
+            }
+        }
+    }
+
+    // MARK: Notifications
+
+    @objc func appStateHasChanged(notification: NSNotification) {
+        let appState = notification.userInfo!["appState"]
```
thisIsTheFoxe:
```suggestion
        let appState = notification.userInfo?["appState"]
```
If you safely unwrap it later anyway, there's no need to force unwrap here :)

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,705 @@
...
+            textField.keyboardType = .phonePad
+            textField.delegate = self
+            textField.tag = kPhoneTextFieldTag
+        }
+        setPhoneAction = UIAlertAction(title: NSLocalizedString("Set", comment: ""), style: .default, handler: { _ in
+            let phoneNumber = setPhoneNumberDialog.textFields?[0].text
+
+            NCAPIController.sharedInstance().setUserProfileField(kUserProfilePhone, withValue: phoneNumber, for: NCDatabaseManager.sharedInstance().activeAccount()) { error, _ in
+                if error != nil {
+                    self.presentPhoneNumberErrorDialog(phoneNumber: phoneNumber!)
```
thisIsTheFoxe: You should only ever use force unwrap when you are 100% sure that the variable must have a value. I'm not sure if that's the case here. I think the best solution here is to handle the case if phoneNumber is optional in `presentPhoneNumberErrorDialog(:)`.
> The exclamation point effectively says, “I know that this optional definitely has a value; please use it.”

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,705 @@
...
+
+let kPhoneTextFieldTag = 99
+
+class SettingsTableViewController: UITableViewController, UITextFieldDelegate {
+
+    var activeUserStatus: NCUserStatus?
+    var readStatusSwitch: UISwitch?
+    var contactSyncSwitch: UISwitch?
+    var setPhoneAction: UIAlertAction?
+    var phoneUtil: NBPhoneNumberUtil?
```
thisIsTheFoxe: If you can be sure that you initialise a variable in `viewDidLoad` and you can be sure that it's never set to `nil` again you could use am implicitly unwrapped variable i.e. `NBPhoneNumberUtil!` which e.g. is also used for `@IBOutlet`s
But I think in this case you could also just initialise it directly, couldn't you?

```swift
class SettingsTableViewController: UITableViewController, UITextFieldDelegate {
  var phoneUtil = NBPhoneNumberUtil()
```

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,705 @@
...
+                } else {
+                    NCContactsManager.sharedInstance().removeStoredContacts()
+                }
+                // Reload to update configuration section footer
+                self.tableView.reloadData()
+        }
+    }
+
+    @objc func readStatusValueChanged(_ sender: Any?) {
+        if let readStatusSwitch = readStatusSwitch {
```
thisIsTheFoxe: You can avoid the _pyramids of doom_ by using `guard`:
```swift
guard let readStatusSwitch = readStatusSwitch else { return }
readStatusSwitch.isEnabled = false
```

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,705 @@
...
+
+        let optionsActionSheet = UIAlertController(title: NSLocalizedString("Open links in", comment: ""), message: nil, preferredStyle: .actionSheet)
+
+        for browser in supportedBrowsers {
+            var isDefaultBrowser = false
+            if let browser = browser as? String {
+                isDefaultBrowser = browser == defaultBrowser
+            }
+            let action = UIAlertAction(title: browser as? String, style: .default) { _ in
+                NCUserDefaults.setDefaultBrowser((browser as? String)!)
```
thisIsTheFoxe: I think `(browser as? String)!` is a bit wired to read. If you absolutely need to force cast just do it one like: `browser as! String`.

thisIsTheFoxe (commented):
Hey,
looks pretty good so far..

Maybe you could be a bit more careful about using forced unwraps (aka `!`). There are some situations where they may make sense, but in general I try to avoid them as much as I can.. :)

NextcloudTalk/SettingsTableViewController.swift
alazarevic5: sure :D

NextcloudTalk/SettingsTableViewController.swift
alazarevic5: changed, now is unwrapped with if let and then in if let block I use safe unwrapped browser.
            if let browser = browser as? String {
                isDefaultBrowser = browser == defaultBrowser
                let action = UIAlertAction(title: browser, style: .default) { _ in
....

.swiftlint.yml
alazarevic5: Yes, I think function with the maximum length is here table view cellForRowAt, because of switch-case for sections... I can revert the rule. 
I have changed the rule, because I got an error, not a warning, when I wanted to add new line in method.

NextcloudTalk/SettingsTableViewController.swift
alazarevic5: Yes

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,696 @@
...
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        self.navigationItem.title = NSLocalizedString("Settings", comment: "")
+        self.navigationController?.navigationBar.titleTextAttributes = [NSAttributedString.Key.foregroundColor: NCAppBranding.themeTextColor()]
+        self.navigationController?.navigationBar.tintColor = NCAppBranding.themeColor()
+        self.tabBarController?.tabBar.tintColor = NCAppBranding.themeColor()
+        self.cancelButton.tintColor = NCAppBranding.themeTextColor()
+
+        phoneUtil = NBPhoneNumberUtil()
```
thisIsTheFoxe: Not sure if this is still needed..? :D

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,696 @@
...
+        return sectionNumber ?? 0
+    }
+
+    func getIndexPathForConfigurationOption(option: ConfigurationSectionOption) -> IndexPath {
+        let section: Int = getSectionForSettingsSection(section: SettingsSection.kSettingsSectionConfiguration)
+        var optionIndexPath: IndexPath = IndexPath(row: 0, section: section)
+
+        let optionRow = getConfigurationSectionOptions().firstIndex(of: option.rawValue)
+
+        if NSNotFound != optionRow {
```
thisIsTheFoxe: Please also check for other occurrences of `NSNotFound`

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,696 @@
...
+        if let failedNumberFormated = try? phoneUtil.format(failedPhoneNumber, numberFormat: NBEPhoneNumberFormat.INTERNATIONAL) {
+            failedPhoneNumberDialog = UIAlertController(
+                       title: alertTitle,
+                       message: alertMessage + "\(failedNumberFormated)",
+                       preferredStyle: .alert)
+        }
+        let retryAction = UIAlertAction(title: NSLocalizedString("Retry", comment: ""), style: .default) { _ in
+            self.presentSetPhoneNumberDialog()
+        }
+        failedPhoneNumberDialog!.addAction(retryAction)
```
thisIsTheFoxe: If `try? phoneUtil.format(failedPhoneNumber` fails this also causes a runtime crash.
Maybe you could just edit your `alertMessage` in the if statement and then create the AlertController after that?

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,696 @@
...
+    }
+
+    // MARK: UITextField delegate
+
+    func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
+        if textField.tag == kPhoneTextFieldTag {
+            let inputPhoneNumber = (textField.text as NSString?)?.replacingCharacters(in: range, with: string)
+            var phoneNumber: NBPhoneNumber?
+            phoneNumber = try? phoneUtil.parse(inputPhoneNumber, defaultRegion: nil)
+            setPhoneAction!.isEnabled = phoneUtil.isValidNumber(phoneNumber)
```
thisIsTheFoxe:
```suggestion
            setPhoneAction?.isEnabled = phoneUtil.isValidNumber(phoneNumber)
```
Optional chaining is an easy way to avoid runtime crashes. Unless you need to make sure `setPhoneAction` actually exists here, in that case you should check for that and handle the error case.

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,696 @@
...
+    override func tableView(_ tableView: UITableView, titleForFooterInSection section: Int) -> String? {
+        let sections = getSettingsSections()
+        let settingsSection = sections[section]
+        if settingsSection == SettingsSection.kSettingsSectionAbout.rawValue {
+            let appName = (Bundle.main.object(forInfoDictionaryKey: "CFBundleDisplayName") as? String)!
+            let appVersion = (Bundle.main.object(forInfoDictionaryKey: "CFBundleShortVersionString") as? String)!
+            return appName + appVersion + copyright
+        }
+        if let activeUserStatus = activeUserStatus {
+            if settingsSection == SettingsSection.kSettingsSectionUserStatus.rawValue && activeUserStatus.status == kUserStatusDND {
```
thisIsTheFoxe: Multiple _conjunctive_ (aka AND) if statements like that should probably be grouped together in one `if` (but not necessarily also in one line). And as mentioned before you can even use `,` instead of `&&` if you want :)

NextcloudTalk/UserAccountTableViewCell.swift
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+import UIKit
+
+let kAccountCellIdentifier: String = "AccountCellIdentifier"
+let kAccountTableViewCellNibName: String = "AccountTableViewCell"
```
thisIsTheFoxe: Again, global variables aren't the worst thing, but I think we can do a little better here.. :)

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,696 @@
...
+    func presentBrowserSelector() {
+        let browserConfIndexPath = self.getIndexPathForConfigurationOption(option: ConfigurationSectionOption.kConfigurationSectionOptionBrowser)
+        let supportedBrowsers = NCSettingsController.sharedInstance().supportedBrowsers!
+        let defaultBrowser = NCUserDefaults.defaultBrowser()
+
+        let optionsActionSheet = UIAlertController(title: NSLocalizedString("Open links in", comment: ""), message: nil, preferredStyle: .actionSheet)
+
+        for browser in supportedBrowsers {
+            var isDefaultBrowser = false
+            if let browser = browser as? String {
```
thisIsTheFoxe: When working on arrays with optional values consider using `compactMap`:
```swift
supportedBrowsers = supportedBrowsers.compactMap({ $0 as? String })
```

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,696 @@
...
+                isDefaultBrowser = browser == defaultBrowser
+                let action = UIAlertAction(title: browser, style: .default) { _ in
+                    NCUserDefaults.setDefaultBrowser(browser)
+                    self.tableView.reloadData()
+                }
+                if isDefaultBrowser {
```
thisIsTheFoxe: This is a total nitpick, but I don't think the variable adds a lot of information compared to just `if browser == defaultBrowser {` :)

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,696 @@
...
+        let privacyCellIdentifier = "PrivacyCellIdentifier"
+        let sourceCodeCellIdentifier = "SourceCodeCellIdentifier"
+        let userStatusCellIdentifier = "UserStatusCellIdentifier"
+
+        let sections = getSettingsSections()
+        let settingsSection = sections[indexPath.section]
+
+        switch settingsSection {
+        case SettingsSection.kSettingsSectionUser.rawValue:
+            let cell = (tableView.dequeueReusableCell(withIdentifier: kUserSettingsCellIdentifier, for: indexPath) as? SettingsUserTableViewCell)!
```
thisIsTheFoxe: `dequeueReusableCell` is also a great fit for using `guard`s:
```swift
guard let cell = tableView.dequeueReusableCell(...) as? SettingsUserTableViewCell else { return UITableViewCell() }
```

.swiftlint.yml
thisIsTheFoxe: I see.. that would be a good candidate for refactoring. The easy solution would be to make functions for each section:
```swift
func userSettingsCell(for indexPath: IndexPath) -> UITableViewCell {
```
But maybe you can find an even better solution e.g. by using protocols for abstraction.

Also, remember that Swift supports type extensions. That can be really helpful in splitting a big ViewController up into smaller parts or even files (e.g. keep only the view life cycle in the original class and use extensions for: UITableView Delegate/DataSource, user interaction, app logic, etc.)

thisIsTheFoxe (changes requested):
Great job! This looks a lot better already.
Feel free to let me know if you have any questions or have other ideas..

NextcloudTalk/SettingsTableViewController.swift
alazarevic5: Yes, I removed this, just missed last time.

NextcloudTalk/SettingsTableViewController.swift
alazarevic5: ok, to check with if let syntax

NextcloudTalk/SettingsTableViewController.swift
alazarevic5: Inside modify message with failed number format and outsite create alert 
if let failedNumberFormated = try? phoneUtil.format(failedPhoneNumber, numberFormat: NBEPhoneNumberFormat.INTERNATIONAL) {
            alertMessage += failedNumberFormated
        }
        failedPhoneNumberDialog = UIAlertController(

NextcloudTalk/SettingsTableViewController.swift
alazarevic5: Changed with ,

.swiftlint.yml
alazarevic5: Yes I put it in extension of class. Method to setup every section cell.

thisIsTheFoxe: That's great, thanks.
Do you still need the rule change then?

thisIsTheFoxe (approved):
Looks good to me 👍

NextcloudTalk/SettingsTableViewController.swift
```diff
@@ -0,0 +1,702 @@
...
+        cell.accountServerLabel.text = accountServer
+        cell.accountImageView.image = NCAPIController.sharedInstance().userProfileImage(for: account, with: CGSize(width: 90, height: 90))
+        cell.accessoryView = nil
+        if account.unreadBadgeNumber > 0 {
+            let badgeView = RoundedNumberView()
+            badgeView.highlightType = kHighlightTypeImportant
+            badgeView.number = account.unreadBadgeNumber
+            cell.accessoryView = badgeView
+        }
+        return cell
```
Ivansss: App crashes when trying to create a inactive account cell. Please check.

Ivansss (changes requested):

Ivansss (approved):

PR #708, opened 2022-01-18
Reduce load on the server when checking call state

Ivansss:
- [x] Use getPeersForCall instead of getParticipantsFromRoom to check call state (it is a lighter request for the server).
- [x] Increased the timer of the request to 5 secs (before 3 secs).

### DIFF
--- a/NextcloudTalk/CallKitManager.m
+++ b/NextcloudTalk/CallKitManager.m
@@ -41,7 +41,7 @@ NSString * const CallKitManagerWantsToUpgradeToVideoCall        = @"CallKitManag
 NSString * const CallKitManagerDidFailRequestingCallTransaction = @"CallKitManagerDidFailRequestingCallTransaction";
 
 NSTimeInterval const kCallKitManagerMaxRingingTimeSeconds       = 45.0;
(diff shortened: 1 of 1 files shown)


NextcloudTalk/CallKitManager.m
```diff
@@ -331,35 +331,35 @@ - (void)checkCallStateForCall:(NSTimer *)timer
...
             [self presentMissedCallNotificationForCall:call];
             [self endCallWithUUID:call.uuid];
             return;
         }
         
+        NSInteger callAPIVersion = [[NCAPIController sharedInstance] callAPIVersionForAccount:account];
+        for (NSMutableDictionary *user in peers) {
+            NSString *userId = [user objectForKey:@"userId"];
+            if (callAPIVersion >= APIv3) {
+                userId = [user objectForKey:@"actorId"];
```
nickvergessen, line 352: need also to check the actorType == users to be on the save side

Ivansss: added the check when call API version >= 3

alazarevic5 (approved):
Tried and code looks good.

PR #713, opened 2022-01-24
User permissions handling

Ivansss:
Adapt calls depending on user permissions:
 - Only request offers if user is sharing audio or video streams
 - Change inCall flags and localMedia depending on user permissions
 - Reconnect to the call if permissions change during a call

### DIFF
--- a/NextcloudTalk/CallViewController.m
+++ b/NextcloudTalk/CallViewController.m
@@ -27,7 +27,6 @@
 #import <WebRTC/RTCEAGLVideoView.h>
 #import <WebRTC/RTCVideoTrack.h>
 
-#import "ARDCaptureController.h"
 #import "DBImageColorPicker.h"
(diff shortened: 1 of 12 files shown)


SystemKeeper:
Just checking: Does this work for Talk < 12? I don’t see a capability check for permissions?

I imagine the following to fail in this case ?
https://github.com/nextcloud/talk-ios/pull/713/commits/24f71f1224264892a5c937e198a17882bd68ae47#diff-4924e88da89128e4862d29acf861ed8f2193a745b7c350546046e978ae62e462R457

Ivansss:
> Just checking: Does this work for Talk < 12? I don’t see a capability check for permissions?

Good point! you are right, the capability check is missing.
There are still some things that might be added to this PR and that one is definitely one that should be added :D

PR #717, opened 2022-01-31
Correct some strings and update localization file

Ivansss:
Correct some strings and update localization file


alazarevic5 (approved):
Looks good!

PR #718, opened 2022-02-01
🧹 Project cleanup

Ivansss:
Remove old classes and pods that are no longer used.
Also renamed some classes to its correct names.


alazarevic5 (approved):
All not needed files are removed. 
Current files are renamed on all places. 
Everything from review was OK

PR #734, opened 2022-02-21
Check signaling state instead of session description type

Ivansss:
Check that signaling state is "HAVE_REMOTE_OFFER" instead of checking session description type.


danxuliu (approved):
Code looks good :+1: (but I have not tested it).

PR #738, opened 2022-03-04
Open conversations

### LINKED ISSUE #640: Join open conversations on mobile
The moderator can open a conversations  for all registered users:
image

In the browser every user can search for this group and join:
image

But on mobile the open conversations are not shown in the search:
image

Ivansss:
This PR adds the ability to search and join "Open conversations"

IMG_0122

Fix #640


SystemKeeper:
Capability added but not used?

Ivansss:
You are right :) It will be used in a following PR when implementing the ability to make a conversation open.
But it could be used already for the search. And do not try to search for open conversations if the capability is not there.

SystemKeeper:
> do not try to search for open conversations if the capability is not there.

Yep, that was my thinking :)

Ivansss:
Added the check here https://github.com/nextcloud/talk-ios/pull/739 ;)

PR #747, opened 2022-03-25
Handle update all participants "inCall" flag message from HPB

### LINKED ISSUE #744: Implement handling for "event.participants.update.all" from HPB
## Steps to reproduce
0. Run HPB
1. Create a conversation
2. Restrict permissions
3. Start a call on web as moderator
4. Join as a normal user in the app
5. As moderator "End meeting for everyone"

### Expected behaviour
The new event of the HPB is handled gracefully:
* All connections are terminated
* The mobile client leaves the call

```json
{"type":"event","event":{"target":"participants","type":"update","update":{"roomid":"5439271893","incall":0,"all":true}}}
```

### Actual behaviour
The event is not handled

### Ref
The spreed PR will only be merged once iOS and Android understand that message and correctly leave the meeting, but currently the "End meeting for all" does not work with more than 50 users, so this should be fixed rather soon:

* https://github.com/nextcloud/spreed/pull/6948
* https://github.com/strukturag/nextcloud-spreed-signaling/pull/208
* https://github.com/nextcloud/talk-android/issues/1881

Ivansss:
Fix #744

### DIFF
--- a/NextcloudTalk/NCCallController.m
+++ b/NextcloudTalk/NCCallController.m
@@ -701,7 +701,20 @@ static NSString * const kNCVideoTrackKind = @"video";
 - (void)externalSignalingController:(NCExternalSignalingController *)externalSignalingController didReceivedParticipantListMessage:(NSDictionary *)participantListMessageDict
 {
     NSLog(@"External participants message received: %@", participantListMessageDict);
+    
     NSArray *usersInRoom = [participantListMessageDict objectForKey:@"users"];
+    
+    // Update for "all" participants
+    if ([[participantListMessageDict objectForKey:@"all"] boolValue]) {
+        // Check if "incall" key exist
+        if ([[participantListMessageDict allKeys] containsObject:@"incall"]) {
+            // Clear usersInRoom array if incall=false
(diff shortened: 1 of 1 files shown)


NextcloudTalk/NCCallController.m
```diff
@@ -701,7 +701,20 @@ - (void)externalSignalingController:(NCExternalSignalingController *)externalSig
 - (void)externalSignalingController:(NCExternalSignalingController *)externalSignalingController didReceivedParticipantListMessage:(NSDictionary *)participantListMessageDict
 {
     NSLog(@"External participants message received: %@", participantListMessageDict);
+    
     NSArray *usersInRoom = [participantListMessageDict objectForKey:@"users"];
```
nickvergessen, line 705: this does not exist in that case btw
So not sure if you need to do
```
if ([[participantListMessageDict allKeys] containsObject:@"users"]) {
```

Ivansss: that check could be added but not really needed.

nickvergessen (approved):
🐘

PR #753, opened 2022-04-13
Store reactions JSON data in database.

Ivansss:
Signed-off-by: Ivan Sein <ivan@nextcloud.com>


SystemKeeper:
Interesting, any reason for this?

Ivansss:
It makes things simpler instead of having to deal with deprecated RLMArray of NCChatReaction :)

SystemKeeper:
I see, haven’t noticed that RLMArray was deprecated 😅 thanks!

PR #800, opened 2022-05-19
Respect the disabled calls setting

### LINKED ISSUE #745: Respect disabled calls setting
https://github.com/nextcloud/spreed/pull/7043 introduces a new setting to disable calls completely.
It is exposed via a capability.
When the setting is off and calls therefor completely disabled the mobile apps should not show the call buttons in the chat view.

nickvergessen:
Fix #745


Ivansss (approved):
Awesome!!
Thanks @nickvergessen :)

PR #801, opened 2022-05-20
l10n: Align source strings

rakekniven:
Signed-off-by: rakekniven <2069590+rakekniven@users.noreply.github.com>


Valdnet (approved):

nickvergessen:
I think the source string needs to be adjusted not the translation, otherwise it is just reverted on the next run?
Please in apps that are not plain php/js always wait for the maintainer to confirm and merge changes

PR #802, opened 2022-05-20
Fix a text string

Valdnet:
Omitted from #801.


rakekniven (approved):
Good catch

PR #811, opened 2022-05-26
Keep parent message when resending a message

### LINKED ISSUE #782: Parent message lost when resending on error
1. Send a chat message as a reply
2. Get a timeout/error
3. Use resend option
4. Message is sent without being a reply

SystemKeeper:
Fixes #782


Ivansss (approved):
Thanks @SystemKeeper :)

PR #814, opened 2022-05-27
Make sure we stay at the bottom, even if multiple reactions are added

### LINKED ISSUE #783: Reaction bottom bar is breaking auto scroll
1. Be in an active chat
2. All kind of messages receive reactions
3. If too many reactions happen without a new message being added in the meantime the view might shift to add the bottom bars so much that the offset is too big and we don't scroll automatically anymore.

SystemKeeper:
Fixes #783


Ivansss:
Can you reproduce reliably #783 @SystemKeeper? 
Because I can't reproduce it 😅
Also, I think that reactions should make the chat scroll just fine because they are received as messages so they will trigger:
https://github.com/nextcloud/talk-ios/blob/master/NextcloudTalk/NCChatViewController.m#L2515

Ivansss (approved):
After being able to reproduce the issue. This PR fixes the issue :D

PR #816, opened 2022-05-27
Debounce WKWebView when running in M1 Mac

### LINKED ISSUE #728: Login via iOS app from M1 Macs not working
I've tried to login to Nextcloud Talk from the app on my M1 Mac to have a standalone talk app instead of a browser tab.
Unfortunately, login doesn't work, claiming wrong username or password. I've just verified this with @Ivansss, login credentials are definitely correct.
We've also tried different nc servers and users, with and w/o 2FA enabled, same issue.
I also gave NC Talk full disc access permissions, same issue.

I'm trying this from my MacBook Pro (13", M1, 2020) using Nextcloud Talk 13.1.1 (1).

Happy to help to debug this but I'd need pointers for how to do so :)

SystemKeeper:
Fixes #728 by implementing a workaround suggested at https://developer.apple.com/forums/thread/696525?answerId=708067022#708067022


Ivansss (approved):
Thanks @SystemKeeper :)

PR #819, opened 2022-05-28
Fix set lobby timer and add translator hint

### LINKED ISSUE #818: Missing translation hint for "Manual"
https://github.com/nextcloud/talk-ios/blob/7f0516ffe527e6e84b712e42197dcfa6b3f88c85/NextcloudTalk/RoomInfoTableViewController.m#L198

This one is translated to "Benutzerhandbuch" in german... I'm not sure if "Manual" is correct in the first place here, but if it is, we should add a hint for translation.

SystemKeeper:
Before this PR:
image

After this PR:
image

Fixes #818 (after understanding what "Manual" referred to...)


Ivansss (approved):
Thanks @SystemKeeper :)

PR #820, opened 2022-05-29
Added login via QR code

### LINKED ISSUE #305: Implement login via QR code
When I choose login with token I can enter name and token. I would expect that I can scan the QR code? But I don't see how. Doesn't start automatically and there is no button I could press..

Do I really have to enter this code manually??

App version 8.1.0

SystemKeeper:
Fixes #305 

image

This is based on the work from the nextcloud/ios project with some minor changes, so I hope it's right to just add me as an author.

There're some checks I'd like to add (like checking if talk is installed, etc.), but it makes sense to do this after decision about #808 was made -> draft. Other than that, adding an account via qrcode should work just fine.


Ivansss (approved):

SystemKeeper:
🤩

PR #821, opened 2022-06-01
🔕 "Silent send" for chat messages

### LINKED ISSUE #787: Support "Silent send for chat messages"
Server ref: https://github.com/nextcloud/spreed/issues/6257

Basic idea: we need a long press / second send button / options menu that allows to send a chat message without triggering notifications.

Ivansss:
Fix #787


NextcloudTalk/NCChatViewController.m
```diff
@@ -1328,7 +1356,7 @@ - (void)didPressResend:(NCChatMessage *)message {
     [self removeUnreadMessagesSeparator];
     
     [self removePermanentlyTemporaryMessage:message];
-    [self sendChatMessage:message.message withParentMessage:message.parent];
+    [self sendChatMessage:message.message withParentMessage:message.parent silently:NO];
```
SystemKeeper, line 1370: Is this obvious to the user? We could just store in NCChatMessage if the message was intended to be silently send and then adjust the resend label and action accordingly?

nickvergessen: Yeah sounds like a valid idea.
I would also expect this. I tried to send silently, just because there was a network interruption or something the resend shouldn't sent with notifications

Ivansss: We will skip this for now until we decide how to store this info on the server side.
At the moment there is no attribute in the room object to know if the message was silently sent.
Once we have this information we could implement what you said @SystemKeeper :)

SystemKeeper: Hm. Why store it on the server side? If we have a temporary message which can be resend, it never hit the server, does it? It would only make sense for message we try to resend, because the call to the server failed. Or am I missing something here?

Ivansss: For this particular case you are right, it doesn't make sense to store it server side since the message won't reach the server.
But the information about a message being sent silently or not is missing on the server at the moment.
Once we decide where this information will be stored in the `message` object, we will be able to modify this attribute in the temporary messages.
I just don't won't to add an attribute (e.g "silent") to the NCChatMessage object (that means modifying the DB schema) until we decide how it will be stored on the server side :)

nickvergessen: Follow up discussion in https://github.com/nextcloud/spreed/issues/8234

PR #824, opened 2022-06-11
Handle "400 Bad Request" on set password

### LINKED ISSUE #823: Handle "400 Bad Request" on set password
Password policy app support has been implemented with https://github.com/nextcloud/spreed/pull/7473
If the password does not match the password policy rules, a `400 Bad Request` will be returned as normal OCS response.
The data contains a `message` attribute which should be shown to the user (it's translated already).

SystemKeeper:
Fixes #823 

image


Ivansss (approved):
Thanks a lot for this PR @SystemKeeper :)

PR #826, opened 2022-06-11
Fix date formatting when searching

SystemKeeper:
When searching for rooms the date formatting is always `dd/MM/yy` instead of the localized version.

Before:
image

After:
image


Ivansss (approved):
Thanks a lot @SystemKeeper :)

PR #829, opened 2022-06-19
Diagnostics

SystemKeeper:
Add some simple diagnostic information to the settings:

image

image

image

image

### DIFF
--- a/NextcloudTalk/DiagnosticsTableViewController.swift
+++ b/NextcloudTalk/DiagnosticsTableViewController.swift
@@ -25,61 +25,60 @@ import Contacts
 import Photos
 import UserNotifications
 
-enum DiagnosticsSections: Int {
-    case kDiagnosticsSectionApp = 0
-    case kDiagnosticsSectionAccount
-    case kDiagnosticsSectionServer
-    case kDiagnosticsSectionTalk
-    case kDiagnosticsSectionSignaling
-    case kDiagnosticsSectionCount
-}
-
-enum AppSections: Int {
-    case kAppSectionName = 0
-    case kAppSectionVersion
-    case kAppSectionAllowNotifications
-    case kAppSectionAllowMicrophoneAccess
-    case kAppSectionAllowCameraAccess
-    case kAppSectionAllowContactsAccess
-    case kAppSectionAllowLocationAccess
-    case kAppSectionAllowPhotoLibraryAccess
-    case kAppSectionOpenSettings
-    case kAppSectionCount
-}
-
-enum AccountSections: Int {
-    case kAccountSectionServer = 0
-    case kAccountSectionUser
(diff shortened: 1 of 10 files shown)


SystemKeeper:
image

SystemKeeper:
* Rebased
* Added talk server version as https://github.com/nextcloud/spreed/pull/7489 is merged
* Capabilities are now shown when talk server version row is selected

image

NextcloudTalk/DiagnosticsTableViewController.swift
```diff
@@ -0,0 +1,654 @@
...
+    case kTalkSectionCount
+}
+
+enum SignalingSections: Int {
+    case kSignalingSectionApiVersion = 0
+    case kSignalingSectionIsExternal
+    case kSignalingSectionStunServers
+    case kSignalingSectionTurnServers
+    case kSignalingSectionCount
+}
```
Ivansss: We could move all this enums inside DiagnosticsTableViewController.
I think there is no need of having them as global variables.

NextcloudTalk/DiagnosticsTableViewController.swift
```diff
@@ -0,0 +1,654 @@
...
+            return NSLocalizedString("Account Diagnostics", comment: "")
+            
+        case DiagnosticsSections.kDiagnosticsSectionServer.rawValue:
+            return NSLocalizedString("Server Diagnostics", comment: "")
+            
+        case DiagnosticsSections.kDiagnosticsSectionTalk.rawValue:
+            return NSLocalizedString("Talk Diagnostics", comment: "")
+            
+        case DiagnosticsSections.kDiagnosticsSectionSignaling.rawValue:
+            return NSLocalizedString("Signaling Diagnostics", comment: "")
```
Ivansss: Since we have "Diagnostics" as title. Maybe we can just have "App", "Account",...as section titles?

NextcloudTalk/DiagnosticsTableViewController.swift
SystemKeeper: Yep, totally right. We should do the same in SettingsTableViewController then: https://github.com/nextcloud/talk-ios/blob/master/NextcloudTalk/SettingsTableViewController.swift

NextcloudTalk/DiagnosticsTableViewController.swift
SystemKeeper: 👍

SystemKeeper:
- [x] Show capabilities when talk server version is not available (Talk < 15)
- [x] Removed internal signaling api (can be optained through capabilities)
- [x] Added todo in #830 for adding external signaling version.  

image

image

SystemKeeper:
image

image

Ivansss (approved):

PR #832, opened 2022-06-23
Remove unneeded iOS version checks

SystemKeeper:
As we now target iOS 12, we can remove any iOS version check < 12.


Ivansss (approved):
Thanks for the needed cleanup :)

PR #848, opened 2022-08-02
📊 Polls support

### LINKED ISSUE #794: Support "📊 Simple polls"
Server ref https://github.com/nextcloud/spreed/issues/7355

Ivansss:
Fix #794


SystemKeeper:
Some things I noticed:

* In some cases we don't receive the number of votes (on open private polls for example) in this case the number of votes should be hidden:
image

  Also on the results view the "Cancel" button should be "Close" in my opinion (we don't cancel anything here anymore).

* Should there be a hint if you already voted on the message view?
image

* Should closed polls be rendered differently in the message view?

* Forwarding und copying is currently allowed (will only forward/copy the title). Is that intended?

* You can't vote nothing, but according to https://github.com/nextcloud/spreed/pull/7621 this should be possible

* UI wise it's not clear if you have 1 or multiple votes. There should be a hint when voting about that

NextcloudTalk/ObjectShareMessageTableViewCell.h
```diff
@@ -0,0 +1,61 @@
+/**
+ * @copyright Copyright (c) 2020 Ivan Sein <ivan@nextcloud.com>
```
SystemKeeper:
```suggestion
 * @copyright Copyright (c) 2022 Ivan Sein <ivan@nextcloud.com>
```

NextcloudTalk/ObjectShareMessageTableViewCell.m
```diff
@@ -0,0 +1,329 @@
+/**
+ * @copyright Copyright (c) 2020 Ivan Sein <ivan@nextcloud.com>
```
SystemKeeper:
```suggestion
 * @copyright Copyright (c) 2022 Ivan Sein <ivan@nextcloud.com>
```

nickvergessen:
> Should closed polls be rendered differently in the message view?

The information is not available until a request was made to the polls API and we would like to avoid that.

Ivansss:
> > Should closed polls be rendered differently in the message view?
> 
> The information is not available until a request was made to the polls API and we would like to avoid that.

I guess we could store some info locally in the DB. But this is out of scope for this PR and for the first version of polls.
Also as @nickvergessen said, this info is not available until the poll info is fetched (e.g. when opening a poll from the app), so the user should at least open the poll once in order to retrieve and show this info in the chat.

SystemKeeper:
Makes sense, that would add too many requests. We do receive the `poll_closed` system message however, which we could use to update the polls appearance. Finding the correct message might be a bit problematic performance-wise, because the poll-id is only available in the json messageParameters.

If that's something that should be achieved, the following ideas come to my mind:
* Search for the poll message in our locally stored messages and update the state (see above)
* Include the original messageId when sending `poll_closed` system message (not sure if possible)
* Store the poll-id locally as the referenceId ("poll-{pollId}" for example) to make it easy to find the message

Not in the scope of this PR though, I just listed what I noticed 😄

Ivansss:
Addressed some of the issues you pointed out @SystemKeeper . The remaining ones will be fix/implemented in a following PR :)

NextcloudTalk/PollVotingView.swift
```diff
@@ -230,11 +230,15 @@ import UIKit
...
     override func tableView(_ tableView: UITableView, titleForHeaderInSection section: Int) -> String? {
-        if section == PollSection.kPollSectionOptions.rawValue && showPollResults {
+        if section == PollSection.kPollSectionOptions.rawValue {
             let votes = poll?.numVoters ?? 0
             let votesString = votes == 1 ? NSLocalizedString("vote", comment: "1 vote in a poll") : NSLocalizedString("votes", comment: "{number} votes in a poll")
             let resultsString = NSLocalizedString("Results", comment: "")
-            return resultsString + " - " + String(votes) + " " + votesString
+            if showPollResults && !showIntermediateResults {
+                return resultsString + " - " + String(votes) + " " + votesString
+            } else if isOwnPoll {
```
SystemKeeper: Do we need to check for moderation permission here as well?

https://github.com/nextcloud/spreed/blob/fcc56adee4b93c4f1e42c1227657986f3d6f276a/lib/Controller/PollController.php#L254-L255

SystemKeeper (approved):

Issue #807, opened 2022-05-24
can not accept call on iphone if nextcloud (talk) is open in browser


ferdiga:
## Steps to reproduce
1. open nextcloud (talk) on Desktop (macbook)
2. open nextcloud talk on iphone
3. ask someone to call you
4. only desktop rings

### Expected behaviour
Step 4 - desktop AND mobile phone should ring
5. select desktop OR mobile phone to answer

### Actual behaviour
Only desktop accepts answer, but conversation runs on iphone 

## Device information

**Device:** iPhone X/iPad Y/…
iphone 12 pro
**iOS version:** 10/…
15.4.1
**Talk version:** 11.1.1
14.0.1
## Server information
Hetzner NC-Cloud service

**Nextcloud version:** (see admin overview page: `/index.php/settings/admin/overview`)
23.0.3
**Talk version:** (see apps admin page: `/index.php/settings/apps`)
13.0.5
**Custom Signaling server configured:** yes/no and version (see talk admin settings: `/index.php/settings/admin/talk#signaling_server`)
No (probably)
**Custom TURN server configured:** yes/no (see talk admin settings: `/index.php/settings/admin/talk#turn_server`)
No (probably)
**Custom STUN server configured:** yes/no (see talk admin settings: `/index.php/settings/admin/talk#stun_server`)
yes
### Server log (data/nextcloud.log)
<details>

```
Insert your server log here
```
</details>

Ivansss:
Hi @ferdiga :)

Do you receive chat messages notifications in your phone? Or you don't receive any notification at all?

> Only desktop accepts answer, but conversation runs on iphone

What do you mean by "conversation runs on iphone"?

ferdiga:
- I receive message information (number of new messages is shown, but not the call) in the iphone , but no call notifications.
- "conversation runs on iphone" I accept on the desktop and talk on the iphone

Ivansss:
Version 14.0.2 fixed some problems regarding push notification subscription process. #806 
Check if you receive push notifications with this version, also you can take a look to this [document](https://github.com/nextcloud/talk-ios/blob/master/docs/notifications.md) for further debugging.

SystemKeeper:
Version 14.0.3 was just released with some additional fixes regarding the push subscription. Can you check if it works for you now @ferdiga ?

PR #852, opened 2022-08-08
Fix saving of push tokens on new devices

### LINKED ISSUE #851: No text notifications on iOS device
## Steps to reproduce
1. users send me a messages
2. sudo -u www-data php /var/www/nextcloud/occ notification:test-push --talk username
3.

### Expected behaviour
I should be getting text notification on my ios devices, but it dose work on my windows desktop client and androind client.
Already rebooted the phone, removed and reinstalled the app sent a sudo -u www-data php /var/www/nextcloud/occ notification:test-push --talk admin no luck

I did do a curl https://push-notifications.nextcloud.com and came back with  404 page not found: based on the doc that a good message. 

### Actual behaviour
Not getting notifications on the IOS devices

## Device information

**Device:** iPhone X/iPad Y/…
iPhone 12 Pro
**iOS version:** 10/…
15.6
**Talk version:** 11.1.1
14.0.2
## Server information

**Nextcloud version:** (see admin overview page: `/index.php/settings/admin/overview`)
24.0.3
**Talk version:** (see apps admin page: `/index.php/settings/apps`)
14.0.3
**Custom Signaling server configured:** yes/no and version (see talk admin settings: `/index.php/settings/admin/talk#signaling_server`)

**Custom TURN server configured:** yes/no (see talk admin settings: `/index.php/settings/admin/talk#turn_server`)
yes
**Custom STUN server configured:** yes/no (see talk admin settings: `/index.php/settings/admin/talk#stun_server`)
yes
### Server log (data/nextcloud.log)
<details>

```
None
```
</details>

SystemKeeper:
On "new" devices (devices where talk-ios was not installed or restored from a backup) the push tokens won't be saved to the keychain, resulting in invalid registration at the push-proxy:
image

Since #806 (to be precise since https://github.com/nextcloud/talk-ios/pull/806/commits/e74adcf60396682b39a08e45368386c553d84add) push tokens and the registration at the server / push proxy only happens when the app is in state `UIApplicationStateActive`. Problem is, that when the app is launched it is first in state `UIApplicationStateInactive`:
> The app is inactive at launch, and then is generally in either an active or background state.
https://developer.apple.com/documentation/uikit/uiapplication/1623003-applicationstate?language=objc

When the tokens are retrieved fast enough, the app is still in state `UIApplicationStateInactive`, therefore the tokens will never get saved to the keychain.  When a new account is added to talk-ios, the subscription process is triggered at https://github.com/nextcloud/talk-ios/blob/eb2abab42bd9a23d9db7f08addae411bf80b99ab/NextcloudTalk/NCSettingsController.m#L173
and the NCKeyChainController will happily return the string `(null) (null)` as `combinedPushToken`, because at this point there's no check if there's actually a stored push token:
https://github.com/nextcloud/talk-ios/blob/eb2abab42bd9a23d9db7f08addae411bf80b99ab/NextcloudTalk/NCKeyChainController.m#L91-L96

This problem does not occur on devices which previously ran an older version of talk-ios and where the push token did not change since it was first started, because the correct push tokens were still in the keychain. Same probably happens when a backup to the same device was restored.

I'm not sure if there's a reason to use `UICKeyChainStore` directly in the `AppDelegate` vs. using `NCKeyChainController`. If there's no reason, we could think about changing this.

Fixes #851 
Related to #807 ?

### DIFF
--- a/NextcloudTalk/AppDelegate.m
+++ b/NextcloudTalk/AppDelegate.m
@@ -175,17 +175,22 @@
 
 - (void)checkForPushNotificationSubscription
 {
-    BOOL isAppActive = [[UIApplication sharedApplication] applicationState] == UIApplicationStateActive;
-    // Subscribe only if both tokens have been generated and app is active (do not try to subscribe when
-    // the app is running in background e.g. when the app is launched due to a VoIP push notification)
-    if (normalPushToken && pushKitToken && isAppActive) {
-        // Store new Normal Push & PushKit tokens in Keychain
-        UICKeyChainStore *keychain = [UICKeyChainStore keyChainStoreWithService:bundleIdentifier accessGroup:groupIdentifier];
-        [NCSettingsController sharedInstance].ncNormalPushToken = normalPushToken;
-        [keychain setString:normalPushToken forKey:kNCNormalPushTokenKey];
-        [NCSettingsController sharedInstance].ncPushKitToken = pushKitToken;
-        [keychain setString:pushKitToken forKey:kNCPushKitTokenKey];
-        
+    if (!normalPushToken || !pushKitToken) {
+        return;
+    }
+
+    // Store new Normal Push & PushKit tokens in Keychain
+    UICKeyChainStore *keychain = [UICKeyChainStore keyChainStoreWithService:bundleIdentifier accessGroup:groupIdentifier];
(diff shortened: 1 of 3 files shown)


Ivansss (approved):
Thanks a lot @SystemKeeper :)

PR #857, opened 2022-08-13
Always call completionHandler in didReceiveNotificationResponse

SystemKeeper:
iOS complains that we do not correctly call the completion handler in `didReceiveNotificationResponse`:

> Warning: UNUserNotificationCenter delegate received call to -userNotificationCenter:didReceiveNotificationResponse:withCompletionHandler: but the completion handler was never called.

Currently the completion handler is called at the end of every called method, so essential it is not different from calling it at the end of `didReceiveNotificationResponse` and making sure it's called in every case. Therefore we can simplify our code and make iOS happy again.


Ivansss (approved):
Makes total sense :) Thanks for the PR @SystemKeeper ;)

PR #817, opened 2022-05-28
Add conversation settings (listable & readonly)

SystemKeeper:
Adds "conversation settings" to the room info:
image

When conversation is open:
image

Fixes #643 ?

PR #858, opened 2022-08-14
Check if the guests app is enabled

SystemKeeper:
Follow-Up to https://github.com/nextcloud/talk-ios/pull/817

Decision was made that the guests-app should expose it's availability on its own capabilities (https://github.com/nextcloud/guests/pull/901). We can now correctly check if the guests-app is available or not and adjust the room options accordingly.

This will be available starting with NC25, earlier versions will only allow to open a conversation to registered users, but not guests. As there was no release with this functionality of talk-ios, I'd argue it's fine to merge as it doesn't break existing functionality.


Ivansss (approved):
Tested and works 👍 
Also it makes sense to show the option only when we can detect if the guest app is enabled (NC25+).
Thanks a lot for this PR @SystemKeeper :)

PR #859, opened 2022-08-18
Fallback for notifications

SystemKeeper:
Background fetch is already implemented in `talk-ios` and periodically checks if there are updates to rooms available. This PR extends the functionality of background fetch and additionally queries the notifications v2 api (https://github.com/nextcloud/notifications/blob/master/docs/ocs-endpoint-v2.md). 
If talk-notifications are available, which were not previously shown by a push notification, we now show a local notification.

Some remarks:
* It's entirely up to iOS to decide when a background fetch is initiated
* This is no replacement for push notifications (especially since there's no guaranteed execution time)
* ETag/If-None-Match is used
* Directly replying to a local notification is not implemented (this can be done, but I don't think it's really worth the work)
* ~Needs to be adjusted after #858 is merged~

How to test:
* Make sure you don't receive push notifications (e.g. use a non-nextcloud bundle id)
* Send a chat message to the user
* Initiate a background-fetch: `Debug -> Simulate Background Fetch`
* Note: Doing a background fetch for the very first time for an account will only update the informations about notifications and not create local notifications on the device. Repeat the steps above to see a local notification appear on the device.

### DIFF
--- a/NextcloudTalk/AppDelegate.m
+++ b/NextcloudTalk/AppDelegate.m
@@ -401,6 +401,17 @@
     [NCUtils log:@"Start performBackgroundFetchWithCompletionHandler"];
 
     dispatch_group_enter(backgroundRefreshGroup);
+    [[NCNotificationController sharedInstance] checkForNewNotificationsWithCompletionBlock:^(NSError *error) {
+        [NCUtils log:@"CompletionHandler checkForNewNotificationsWithCompletionBlock"];
+
+        if (error) {
+            errorOccurred = YES;
+        }
+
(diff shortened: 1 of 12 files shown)


Ivansss:
We can now rebase this PR since #858 has been merged ;)

SystemKeeper:
Before merging, we might want to think about limiting the number of notifications shown to the user... Maybe add a local notification like "You received X notifications, only the first Y will be shown. Please open the app." or something.

Ivansss (approved):

PR #861, opened 2022-08-25
Plurals support

Ivansss:
Use stringdict file for plurals support in different languages.

- [x] Create new project in transifex https://github.com/nextcloud/docker-ci/issues/405

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -8,6 +8,13 @@ source_lang = en
 type = STRINGS
 lang_map = pt_BR:pt-BR,zh_CN:zh-Hans,fi_FI:fi-FI,es_MX:es-MX,nb_NO:nb-NO,cs_CZ:cs,en_GB:en-GB,es_AR:es-AR,sk_SK:sk-SK,hu_HU:hu,ka_GE:ka-GE,zh_TW:zh-Hant-TW,es_CL:es-CL,es_CO:es-CO,es_CR:es-CR,es_DO:es-DO,es_EC:es-EC,es_GT:es-GT,es_HN:es-HN,es_NI:es-NI,es_PA:es-PA,es_PE:es-PE,es_PR:es-PR,es_PY:es-PY,es_SV:es-SV,es_UY:es-UY,es_419:es-419,pt_PT:pt-PT,ja_JP:ja
 
+[nextcloud.talk-ios-plurals]
+file_filter = NextcloudTalk/<lang>.lproj/Localizable.stringsdict
+source_file = NextcloudTalk/en.lproj/Localizable.stringsdict
+source_lang = en
+type = STRINGSDICT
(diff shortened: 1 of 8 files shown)


nickvergessen:
Can we merge this to test it?

.tx/config
```diff
@@ -8,6 +8,13 @@ source_lang = en
 type = STRINGS
 lang_map = pt_BR:pt-BR,zh_CN:zh-Hans,fi_FI:fi-FI,es_MX:es-MX,nb_NO:nb-NO,cs_CZ:cs,en_GB:en-GB,es_AR:es-AR,sk_SK:sk-SK,hu_HU:hu,ka_GE:ka-GE,zh_TW:zh-Hant-TW,es_CL:es-CL,es_CO:es-CO,es_CR:es-CR,es_DO:es-DO,es_EC:es-EC,es_GT:es-GT,es_HN:es-HN,es_NI:es-NI,es_PA:es-PA,es_PE:es-PE,es_PR:es-PR,es_PY:es-PY,es_SV:es-SV,es_UY:es-UY,es_419:es-419,pt_PT:pt-PT,ja_JP:ja
 
+[nextcloud.talk-ios-plurals]
```
nickvergessen, line 11: @rakekniven new resource for iOS talk plurals. Once this is confirmed working (and it seems to work at the moment), we will also do a -plurals for the files app.
Can you add it to your forum post list?

NextcloudTalk/PollVotingView.swift
```diff
@@ -234,12 +234,12 @@ import UIKit
     override func tableView(_ tableView: UITableView, titleForHeaderInSection section: Int) -> String? {
         if section == PollSection.kPollSectionOptions.rawValue {
             let votes = poll?.numVoters ?? 0
-            let votesString = votes == 1 ? NSLocalizedString("vote", comment: "1 vote in a poll") : NSLocalizedString("votes", comment: "{number} votes in a poll")
-            let resultsString = NSLocalizedString("Results", comment: "")
+            let votesString = String(format: NSLocalizedString("%d votes", comment: "Votes in a poll"), votes)
```
SystemKeeper, line 237: Shouldn’t `localizedStringWithFormat` be used here?

nickvergessen: Ivan will fix the stuff today. We mostly merged it to test if the syncing to transifex was working or needed more work somewhere.

Ivansss: `localizedStringWithFormat` definition says:

```
Returns a string created by using a given format string as a template into which the remaining argument values are substituted according to the current locale.
```

So yes, let's change it in case the argument needs to be localized for some languages ;)

.tx/config
rakekniven: > @rakekniven new resource for iOS talk plurals. Once this is confirmed working (and it seems to work at the moment), we will also do a -plurals for the files app. Can you add it to your forum post list?

Ok, will add it. Thanks for notifying.
btw. the forum post is a wiki and everyone can modify it.

PR #864, opened 2022-08-31
Only use dav endpoint

Ivansss:
Only use `/remote.php/dav` endpoint when interacting with files.

### DIFF
--- a/NextcloudTalk/NCAPIController.h
+++ b/NextcloudTalk/NCAPIController.h
@@ -99,6 +99,7 @@ typedef void (^UnsubscribeToNextcloudServerCompletionBlock)(NSError *error);
 typedef void (^SubscribeToPushProxyCompletionBlock)(NSError *error);
 typedef void (^UnsubscribeToPushProxyCompletionBlock)(NSError *error);
 
+extern NSString *const davEndpoint;
 extern NSInteger const APIv1;
 extern NSInteger const APIv2;
 extern NSInteger const APIv3;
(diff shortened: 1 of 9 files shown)


NextcloudTalk/NCAPIController.h
```diff
@@ -99,6 +99,7 @@ typedef void (^UnsubscribeToNextcloudServerCompletionBlock)(NSError *error);
 typedef void (^SubscribeToPushProxyCompletionBlock)(NSError *error);
 typedef void (^UnsubscribeToPushProxyCompletionBlock)(NSError *error);
 
+extern NSString *const davEndpoint;
```
SystemKeeper: Doesn't need to be public as it's only used inside of NCAPIController

NextcloudTalk/NCAPIController.m
```diff
@@ -38,6 +38,7 @@
 NSInteger const APIv3                       = 3;
 NSInteger const APIv4                       = 4;
 
+NSString * const davEndpoint                = @"/remote.php/dav";
```
SystemKeeper: Might consider renaming to `kNCDavEndpoint` to be inline with the other consts

SystemKeeper (commented):
Any reason to still keep the `webDAVRoot`-property around? Only place where `webDAVRoot` is still used is in the diagnostics view, so we could remove it?!

SystemKeeper (approved):

PR #867, opened 2022-09-08
Background improvements

SystemKeeper:
This PR improves the logic when doing a background-fetch and also adds logging for the same. There is room for improvements here, but to further diagnose crashes it should be a good starting point.
This PR also allows access to the documents directory from iTunes and from the iOS files app - this is fine for the next beta, but should be revisited before release.


SystemKeeper:
By default apple does not allow access to files while in the background, so the last commit moves the logs to a dedicated folder with `NSFileProtectionNone`.

Ivansss (approved):

PR #873, opened 2022-09-15
Fix create new conversation issue

Ivansss:
When the search controller is active without any text in it, you could not select "Group" or "Public" converastion options.
This PR fixes the search controller check and allows to click on those options.


SystemKeeper (approved):
Makes sense 👍

PR #875, opened 2022-09-16
Hide empty sections in chat

SystemKeeper:
Before:
image

After:
image

### DIFF
--- a/NextcloudTalk/NCChatViewController.m
+++ b/NextcloudTalk/NCChatViewController.m
@@ -3348,6 +3348,23 @@ NSString * const NCChatViewControllerTalkToUserNotification = @"NCChatViewContro
     if ([tableView isEqual:self.autoCompletionView]) {
         return 0;
     }
+
+    NSDate *date = [_dateSections objectAtIndex:section];
+    NSMutableArray *messages = [_messages objectForKey:date];
+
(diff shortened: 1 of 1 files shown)


NextcloudTalk/NCChatViewController.m
```diff
@@ -3348,6 +3348,23 @@ - (CGFloat)tableView:(UITableView *)tableView heightForHeaderInSection:(NSIntege
...
+    NSMutableArray *messages = [_messages objectForKey:date];
+
+    NSUInteger idx = [messages indexOfObjectPassingTest:^BOOL(NCChatMessage *message, NSUInteger idx, BOOL * _Nonnull stop) {
+        if (![message isUpdateMessage]) {
+            *stop = YES;
+            return YES;
+        }
+
+        return NO;
+    }];
```
Ivansss: Nice idea to hide empty sections ;)
Maybe we could reuse the function `newMessagesContainVisibleMessages()` and rename it to something like `messagesContainVisibleMessages()`, what do you think? :)

Ivansss (changes requested):

NextcloudTalk/NCChatViewController.m
SystemKeeper: Oh you’re right, good idea  👍 I’ll update the PR

Ivansss (approved):

PR #878, opened 2022-09-20
Hide create poll option in one2one rooms

Ivansss:
Ref: https://github.com/nextcloud/spreed/issues/7934

### DIFF
--- a/NextcloudTalk/NCChatViewController.m
+++ b/NextcloudTalk/NCChatViewController.m
@@ -1273,7 +1273,7 @@ NSString * const NCChatViewControllerTalkToUserNotification = @"NCChatViewContro
         [optionsActionSheet addAction:cameraAction];
     }
     [optionsActionSheet addAction:photoLibraryAction];
-    if ([[NCDatabaseManager sharedInstance] serverHasTalkCapability:kCapabilityTalkPolls]) {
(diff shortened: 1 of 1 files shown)


NextcloudTalk/NCChatViewController.m
```diff
@@ -1273,7 +1273,7 @@ - (void)presentAttachmentsOptions
         [optionsActionSheet addAction:cameraAction];
     }
     [optionsActionSheet addAction:photoLibraryAction];
-    if ([[NCDatabaseManager sharedInstance] serverHasTalkCapability:kCapabilityTalkPolls]) {
+    if ([[NCDatabaseManager sharedInstance] serverHasTalkCapability:kCapabilityTalkPolls] && _room.type != kNCRoomTypeOneToOne) {
```
nickvergessen, line 1276: Are you checking chat permissions? (similar as for other attachments)

nickvergessen (approved):

NextcloudTalk/NCChatViewController.m
Ivansss: if the user doesn't have chat permission the input field (and the attachments button) is hidden.
So the user won't be able to select any attachment option.

PR #880, opened 2022-09-21
Handle "Enable SIP without PIN requirement"

### LINKED ISSUE #772: Respect "Allow to enable SIP without individual pin requirement"
Server ref https://github.com/nextcloud/spreed/pull/6563

`sipEnabled` can now be 0-2 instead of 0 or 1
The posting endpoint is unchanged and allows this already but basically we need either a drop down or 2 checkboxes now.

Screenshot of the web:
Bildschirmfoto vom 2022-05-03 09-30-21

Ivansss:
Fix #772


NextcloudTalk/RoomInfoTableViewController.m
```diff
@@ -1419,9 +1428,19 @@ - (void)sipValueChanged:(id)sender
 {
     _sipSwitch.enabled = NO;
```
SystemKeeper, line 1429: `_sipNoPINSwitch.enabled = NO;`

Should be disabled here as well, like it is with listable.

SystemKeeper (changes requested):

NextcloudTalk/RoomInfoTableViewController.m
Ivansss: Done ;)

SystemKeeper (approved):
👍

PR #883, opened 2022-09-22
Fix reaction on mac

### LINKED ISSUE #837: Unable to add reactions on Mac
## Steps to reproduce
1. Right click on a message 
2. Select "Add reaction"

### Expected behaviour
Emoji picker should open

### Actual behaviour
Nothing happens. Answering works fine.

https://user-images.githubusercontent.com/1580193/176171532-ea0cebcf-25c2-4613-9664-3b2c3b8315c7.mp4

SystemKeeper:
Fixes #837 

The way we invoke the emoji keyboard in iOS does not work when running on a Mac. In this case we need to open the emoji picker built into Mac OS. This can be done by accessing `NSApplication` which by default is not accessible in an iOS application. Therefore the `Dynamic` library was added to access the needed methods. 
To show the emoji picker always at the correct location, we need to remove and then re-add the `emojiTextField`. Otherwise the location will be correct for the first time we open the picker, but depending on how the second row was selected, it might stay at the old location (although still targeting the new row).

There's still a small issue left (probably not happening a lot in real world usage): 
* Choose "Add reaction" on row 1
* Leave emoji picker open
* Do a right click on row 2 and select "Add reaction"
* The picker will not be visible anymore, but shows, after another click

### DIFF
--- a/NextcloudTalk.xcodeproj/project.pbxproj
+++ b/NextcloudTalk.xcodeproj/project.pbxproj
@@ -8,6 +8,8 @@
 
 /* Begin PBXBuildFile section */
 		1F3D3B22255F109E00230DAE /* BarButtonItemWithActivity.m in Sources */ = {isa = PBXBuildFile; fileRef = 1F3D3B20255F109E00230DAE /* BarButtonItemWithActivity.m */; };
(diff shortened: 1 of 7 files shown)


SystemKeeper:
Bildschirmfoto 2022-09-23 um 08 34 29

Ivansss (approved):

PR #889, opened 2022-09-26
Save camera flash mode in user defaults

### LINKED ISSUE #885: Save flash setting for camera attachment option
Currently when I want to quickly send a picture, I tap the attachments clip icon and choose "Camera". This opens a camera view which always has the flash set to auto.

It would be nice if the previous flash setting (on/off/auto) was remembered, for example I usually shoot without the flash and every time I forget that this app activates it so I have to re-take the picture after deactivating it.

Always remembering to manually turn off the flash, but it's more of an annoyance rather than an alternative.

**Additional context**
Flash settings in the camera view (after tapping on the flash icon at the top):

Ivansss:
Fix #885


SystemKeeper (approved):
Tested and works 👍

PR #894, opened 2022-09-30
Push notification subscription enhancements

### LINKED ISSUE #855: Push notifications review
While doing some research for #852 I noticed a few things, which might be worth discussing:

* The property `pushNotificationSubscribed` does not have a benefit right now:
  https://github.com/nextcloud/talk-ios/blob/eb2abab42bd9a23d9db7f08addae411bf80b99ab/NextcloudTalk/TalkAccount.h#L46
  it's currently resetted in `AppDelegate`, but never actually read:
  https://github.com/nextcloud/talk-ios/blob/eb2abab42bd9a23d9db7f08addae411bf80b99ab/NextcloudTalk/AppDelegate.m#L189-L195
  Taking https://github.com/nextcloud/talk-ios/issues/691 into account, I'd suggest we remove the boolean `pushNotificationSubscribed` and instead add a timestamp `lastSuccessfulPushSubscription`. This way we can determine when to subscribe again (~24h), for example in a background refresh cycle or in `applicationDidBecomeActive` (but still force the subscription when the app is launched). It also allows us to show a hint in case the subscription process fails for multiple days (could be added to #830)

* The error handling for generating the RSA keys is wrong:
https://github.com/nextcloud/talk-ios/blob/eb2abab42bd9a23d9db7f08addae411bf80b99ab/NextcloudTalk/NCSettingsController.m#L585-L588
Only thing that happens in the `generateRSAKey` method is a cleanup in case of an error, but there will be no NSError returned as far as I can see:
https://github.com/nextcloud/talk-ios/blob/eb2abab42bd9a23d9db7f08addae411bf80b99ab/NextcloudTalk/NCSettingsController.m#L634-L644
Also the checked return values are wrong. From the documentation ([1](https://www.openssl.org/docs/man1.1.1/man3/RSA_generate_key_ex.html) & [2](https://www.openssl.org/docs/man1.1.1/man3/BN_set_word.html)):
   ```
   BN_one() and BN_set_word() return 1 on success, 0 otherwise.
   RSA_generate_key_ex() returns 1 on success or 0 on error.
   ```
   So we should not check for `< 0` here.

  There's also no error handling in `generatePushNotificationsKeyPairForAccountId`. I suggest we either do proper error handling in both cases or remove the wrong error handling all together. 

* It does not look like `NCSettingsController` still needs access to the push tokens, but still they're set in `AppDelegate`:
  https://github.com/nextcloud/talk-ios/blob/eb2abab42bd9a23d9db7f08addae411bf80b99ab/NextcloudTalk/AppDelegate.m#L183-L187
  I suggest we remove those properties from `NCSettingsController` and use `NCKeyChainController` here directly.
  (Also I'd like to suggest to remove the legacy keychain migration code. Multi-Account support and therefore migration from keychain to database was introduced in version 8, around 3 years ago. I think we can simplify this and don't keep a migration from version < 8 to 15..)

Ivansss:
- [x] Use the same key pair when subscribing for push notifications.
- [x] Show last subscription date in diagnostics view.
- [x] Fix error handling when generating a key pair.
- [x] Remove legacy keychain migration code.

Fix #855


SystemKeeper (approved):

NextcloudTalk/NCSettingsController.m
```diff
@@ -555,24 +481,34 @@ - (BOOL)isGuestsAppEnabled
...
+        keyPair = [[NCPushNotificationKeyPair alloc] init];
+        keyPair.publicKey = pushNotificationPublicKey;
+        keyPair.privateKey = pushNotificationPrivateKey;
+    } else {
+        keyPair = [self generatePushNotificationsKeyPairForAccountId:accountId];
+    }
+    
     if (!keyPair) {
-        NSLog(@"Error while subscribing: Unable to generate push notifications key pair.");
+        [NCUtils log:@"Error while subscribing: Unable to generate push notifications key pair."];
```
SystemKeeper, line 497: Differentiate between unable to generate and no keyPair available?

NextcloudTalk/NCSettingsController.m
```diff
@@ -594,27 +530,26 @@ - (void)subscribeForPushNotificationsForAccountId:(NSString *)accountId
                     NSPredicate *query = [NSPredicate predicateWithFormat:@"accountId = %@", accountId];
                     TalkAccount *managedAccount = [TalkAccount objectsWithPredicate:query].firstObject;
                     managedAccount.pushNotificationPublicKey = keyPair.publicKey;
```
SystemKeeper: Can be removed as it's not used anymore?!

NextcloudTalk/NCSettingsController.m
Ivansss: Left it like that as discussed.

NextcloudTalk/NCSettingsController.m
Ivansss: Removed ;)

PR #907, opened 2022-10-05
Update Realm lib to version 10.30.0.

Ivansss:
Fixes "Warning about use of private API when uploading app with Xcode 14"
Ref: https://github.com/realm/realm-swift/issues/7945


SystemKeeper (approved):
Couldn't spot any difference

PR #912, opened 2022-10-07
Fixes in mentions

### LINKED ISSUE #910: Mentions are broken when re-sending a failed message

Ivansss:
- [x] Include mentions when resending a failed message.
- [x] Highlight mentions in temporary messages.
- [x] Do not delete all mentions from same user/room when deleting a single one.

Fix #910


SystemKeeper (approved):
Tested and works 👍

PR #913, opened 2022-10-10
Implement iOS 16 workaround for answering video calls

### LINKED ISSUE #903: Calls can't be answered on iPhones with IOS16
## Steps to reproduce
1. Have Nextcloud-IOS installed on an iPhone with IOS16.0.2
2. Start a video call from a different account
3. On the iPhone, the call appears on the lock screen. Swipe to answer the call.
4. iPhone is doing FaceID
5. iPhone shows the spring board, the NC app does not open and handle the call

### Expected behaviour
The incoming call should be answered.

### Actual behaviour
The talk app is not opened, the call is not answered.

In case the iPhone is unlocked, the call banner appears on top of the screen. In this scenario, pressing the blue "answer" button *does* open the NC talk app, but it's not handling the call. Also here the call can't be answered.

## Device information

**Device:** iPhone 11 Pro Max

**iOS version:** 16.0.2

**Talk version:** 14.0.3

## Server information

**Nextcloud version:** 24.0.5

**Talk version:** 14.0.5

**Custom Signaling server configured:** yes, 1.0.0 (33dc5a554b496bd253bedbecc6755de514a64867)

**Custom TURN server configured:** yes

**Custom STUN server configured:** yes

### Additional info

Works on an iPhone 8 Plus with IOS 15.5 with the same account and app version.

SystemKeeper:
Fixes #903 

Checks if the device is actually unlocked before fulfilling the CallKit action (as suggested in apples developer forum). Otherwise the user either drops to springboard after unlocking a device or lands in the CallKit screen. This mainly affects video calls.

How to test:
- Move Nextcloud Talk to the background and lock the device
- Start a video call from a second device
- Slide to answer the call -> the phones asks to unlock 
- After unlock, Nextcloud Talk should be open with video call

### DIFF
--- a/NextcloudTalk/CallKitManager.m
+++ b/NextcloudTalk/CallKitManager.m
@@ -41,6 +41,7 @@ NSString * const CallKitManagerWantsToUpgradeToVideoCall        = @"CallKitManag
 NSString * const CallKitManagerDidFailRequestingCallTransaction = @"CallKitManagerDidFailRequestingCallTransaction";
 
 NSTimeInterval const kCallKitManagerMaxRingingTimeSeconds       = 45.0;
+NSTimeInterval const kCallKitManagerMaxWaitToUnlockSeconds      = 45.0;
 NSTimeInterval const kCallKitManagerCheckCallStateEverySeconds  = 5.0;
 
 @interface CallKitManager () <CXProviderDelegate>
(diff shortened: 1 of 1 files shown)


Ivansss (changes requested):
We should check the iOS version (16) and if the call has video, since the "problem" seems to only occur in this scenario.

Also, delaying the answer until the phone is unlocked doesn't seem to be a good workaround.  There might be scenarios where the user don't/can't unlock the phone but still wants to pick up a videocall and start the conversation just with voice while the phone is locked.

Finally, since this seems to be an iOS change/bug, I would just wait for a fix from Apple and do not try to workaround this issue.

PR #915, opened 2022-10-12
Use reference provider to retrieve details of a shared deck card

SystemKeeper:
Before:

After:

The changes in `NCSettingsController` are not really necessary at the moment (*), for completeness it makes sense, but could also be removed... I'm fine either way :-)

*) When we use `NCSettingsController` in `NCChatMessage` we need to introduce additional dependencies to notification service extension, which I would prefer not to do.


SystemKeeper:
For testing with the latest stable25 this commit is needed (not yet released as far as I can see): https://github.com/nextcloud/deck/commit/ed5df26704b4754370f0bfa5ea647e628c8e47f0

Ivansss (approved):

PR #919, opened 2022-10-13
Remove SocketRocket and use iOS native web sockets

SystemKeeper:
As we don't target iOS < 13 anymore, we can switch to the native web socket implementation of iOS. This removes SocketRocket as a dependency and should eliminate crashes surrounding web socket connections.

Corresponding apple documentation: https://developer.apple.com/documentation/foundation/nsurlsessionwebsockettask


SystemKeeper:
Hm, I just noticed that SocketRocket had `allowsUntrustedSSLCertificates` set... I guess that this will now fail for the new implementation. I'll take a look.

Ivansss (approved):

PR #928, opened 2022-10-17
Allow message expiration in one2one rooms

Ivansss:
Ref: https://github.com/nextcloud/spreed/pull/8166


SystemKeeper (approved):
Tested and works. Changes ordering of the sections, so conversation settings are now above guest settings. I don't think this will be a problem?!

Issue #242, opened 2019-07-10
macOS application with Catalyst


FramboisePi:
Hello,

As said here in https://github.com/nextcloud/spreed/issues/1100, it would be really useful to have a macOS application for Nextcloud Talk.

Apple recently introduced Project Catalyst which allows to port iOS app to macOS easily. 
https://developer.apple.com/ipad-apps-for-mac/

I don’t know if it would be possible to make a client this way. If it is the case, I think it would be a great improvement for Talk.

Ivansss:
> Apple recently introduced Project Catalyst which allows to port iOS app to macOS easily.
> https://developer.apple.com/ipad-apps-for-mac/

We were really excited when Apple did that announcement :)
We will definitely try to port this iOS app to a macOS app.

I will post updates about this process in this issue.

tiiiecherle:
Are there any news or developments on this? We would really like to use spreed/talk in production but the desktop app for mac is an essential requirement.

Has anyone managed to build natively, with catalyst or nativefier? Or any other way or integration for a mac desktop app.

Thanks to the nextcloud and spreed/talk team, amazing work ;)

I'll be thankful for any help and would happily try a beta or test version.

Ivansss:
We didn't start implementing this, but we will try it at some point.
I will update this issue when we have news :)

f1yankees:
+1 on this, would be very helpful in these times!

aniqueta:
Agreed. @Ivansss This would be a more performant option for Macs than using the web version.

BJKle:
@Ivansss any news here? I've been using the IOS version on my M1 system but a native app would be really great.

BJKle:
@Ivansss I have to ask again, because the new version 13 crashes and can't be used anymore. :-(

Ivansss:
> any news here? I've been using the IOS version on my M1 system but a native app would be really great.

A specific UI for the iPad is in the roadmap :)

> I have to ask again, because the new version 13 crashes and can't be used anymore. :-(

Could you paste here any logs from the crash?

SystemKeeper:
```
(
	0   CoreFoundation                      0x000000019b583838 __exceptionPreprocess + 240
	1   libobjc.A.dylib                     0x000000019b2ad0a8 objc_exception_throw + 60
	2   CoreFoundation                      0x000000019b614694 -[NSObject(NSObject) __retain_OA] + 0
[...]

PR #936, opened 2022-10-20
Mac Catalyst support

SystemKeeper:
This is/was an attempt to add support for Mac catalyst. This PR is not finished or polished in any way, it should be considered as a PoC.

Building and running as a catalyst app works fine. The following problems are unsolved:
* Push notifications are not working (this might be something that needs to be fixed by Nextcloud GmbH and not in this code)
* Calls are not connecting properly. At this point I'm not sure if this is a problem on our end, or a problem with WebRTC build for catalyst as it looks like Google is not putting much effort into supporting catalyst.
* One build script (the one about removing unused architectures) sometimes fails to run. I skipped this run script in this branch.

At this point I don't think there will be much more effort in this PR and I'm not sure that catalyst support is actually the way to go for this repo.

Ref #242 
Updated instructions for building WebRTC for catalyst: https://github.com/nextcloud-releases/talk-clients-webrtc

### DIFF
--- a/NextcloudTalk.xcodeproj/project.pbxproj
+++ b/NextcloudTalk.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 52;
+	objectVersion = 55;
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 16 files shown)


SystemKeeper:
There are no plans to work on this any further. Development resources are directed to towards https://github.com/nextcloud/talk-desktop

PR #940, opened 2022-10-20
Report all calls to CallKit

SystemKeeper:
Since iOS 13 we're forced to report all received VoIP push notifications to CallKit. Failing in doing so multiple times restricts the app to only receive VoIP push notification when the app is running, but not in the background anymore.
Under normal circumstances this is not a problem, because we want to show calls. A problem arises if the device was previously used with talk-ios, but the account was not properly removed (talk-ios was removed, but not the account). This results in the server sending push notifications to the device with an unknown encryption key. In that case we can't decrypt the message and are unable to properly show the call notification.

This PR reports all calls to CallKit, even those which we are unable to properly decrypt. In that case "Old account" will be displayed as caller id and a local notification will be posted. Additionally we remember that we received such a problematic notification and show information in the settings controller.

image

image

### DIFF
--- a/NextcloudTalk/NCSettingsController.m
+++ b/NextcloudTalk/NCSettingsController.m
@@ -83,8 +83,10 @@ NSString * const kUserProfileScopePublished     = @"v2-published";
 
 NSInteger const kDefaultChatMaxLength           = 1000;
 
-NSString * const kPreferredFileSorting  = @"preferredFileSorting";
-NSString * const kContactSyncEnabled  = @"contactSyncEnabled";
+NSString * const kPreferredFileSorting          = @"preferredFileSorting";
+NSString * const kContactSyncEnabled            = @"contactSyncEnabled";
+
+NSString * const kDidReceiveCallsFromOldAccount = @"receivedCallsFromOldAccount";
 
 + (NCSettingsController *)sharedInstance
 {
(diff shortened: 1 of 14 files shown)


NextcloudTalk/NCSettingsController.m
```diff
@@ -246,6 +248,18 @@ - (void)setContactSync:(BOOL)enabled
...
+    BOOL didReceiveCallsFromOldAccount = [[[NSUserDefaults standardUserDefaults] objectForKey:kDidReceiveCallsFromOldAccount] boolValue];
+
+    return didReceiveCallsFromOldAccount;
+}
+
+- (void)setDidReceiveCallsFromOldAccount:(BOOL)receivedOldCalls
+{
+    [[NSUserDefaults standardUserDefaults] setObject:@(receivedOldCalls) forKey:kDidReceiveCallsFromOldAccount];
+}
+
```
Ivansss, line 262: Maybe we could move this to NCUserDefaults? :)

Ivansss (changes requested):

NextcloudTalk/NCSettingsController.m
Ivansss: We'll handle it in a follow-up PR

Ivansss (approved):

PR #957, opened 2022-10-26
i18n: Separate words

Valdnet:


Ivansss (approved):
Thanks a lot @Valdnet :)

PR #955, opened 2022-10-26
Better in chat media preview

e-caste:
Close #776 

This PR is a draft: some constraints are breaking (even though visually the app is working), and a couple TODOs need to be done (I'm no iOS developer, so I don't want to try submitting some very bad code to do the couple of missing things).

It would also be nice to load higher resolution image previews from the server, since now they are bigger.

Current state:
IMG_A9A1DF448FFD-1

PR #963, opened 2022-10-29
Prepend emoji to body text of messages containing files

e-caste:
this should apply to:
- message previews in the chat list view ✅
- messages inside a chat ✅
- notifications (I can't check since I can't receive notifications with my developer certificate)

This should help identify the nature of a message in lieu of a media preview.

Simulator Screen Shot - iPhone 14 Pro - 2022-10-29 at 12 27 19

Simulator Screen Shot - iPhone 14 Pro - 2022-10-29 at 13 03 47

### DIFF
--- a/NextcloudTalk/NCChatMessage.m
+++ b/NextcloudTalk/NCChatMessage.m
@@ -24,6 +24,7 @@
 
 #import "NCAPIController.h"
 #import "NCAppBranding.h"
+#import "NCUtils.h"
 #import "NextcloudTalk-Swift.h"
 
 NSInteger const kChatMessageGroupTimeDifference = 30;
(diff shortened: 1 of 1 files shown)


e-caste:
Hey @SystemKeeper @Ivansss what do you think about this? I know it seems silly, but in my experience it should be a good enough UX improvement to be able to quickly recognise the type of attachment directly from the message text. Let me know if you need any change to be able to merge this, by the way I was thinking about adding one method per MIME type in `NCUtils` like the two I have added in #955.  
Also as I mentioned above I can't test if this is the same text that is received in the notifications but I hope so, otherwise please tell me what I need to add.  
Should I also handle the maximum character limit in any way? Cause if I add `"🅱️ "` it takes 2 characters away from the limit, one for the emoji and one for the space (even 3 in cases of emojis with ligatures such as person/hand+specific color e.g. 👌🏼).

SystemKeeper:
Hey @e-caste 
Sorry for not replying earlier :-). In general I like the idea. It's nice to see a poll or contact in the conversation list for example (also in push notifications). I do think however that it is kind of redundant in the chat itself, because if there's a shared contact (or map, or file, etc)., we already have a big icon, so I don't think we need another one. (Also I'm not sure about file sharing, because a folder does not look right for a .docx file, but that might be fine-tuning).
Actually I did create #976 with your PR in mind 😉 It would allow us to separate the message used in the chat and in other places. What do you think about that?
I can talk with Ivan about it next week. We first wanted to merge your preview enhancements, but other things popped up 😄

e-caste:
@SystemKeeper Yeah makes sense! It would also allow us to expand on this idea in the future, like for example Telegram shows a very small thumbnail of the actual attachment instead of an emoji for its kind (though it showed an emoji in previous versions).

SystemKeeper:
Yea, that sounds like a nice idea! I think that makes more sense than emojis, at least for files and images. For "objects" (poll, map, contact, etc). I think emojis are totally fine

SystemKeeper:
Hey @e-caste 
just wanted to check with you, would you want to adjust this PR with the new situation (#976 merged) or should we take a look at some point? :-)

e-caste:
Hey @SystemKeeper thanks for asking. I don't think I'll have time in the near future for this, but you can absolutely add more commits to this PR if you need to update it to be able to merge it

SystemKeeper:
Continuation in https://github.com/nextcloud/talk-ios/pull/1805

PR #966, opened 2022-10-31
Don't deliver local notifications when user status is DND

### LINKED ISSUE #962: Respect "DND" when fetching notifications in a background task
Check for user status before presenting notifications that have been fetched in a background task.

SystemKeeper:
Fixes #962 
Ref https://github.com/nextcloud/notifications/pull/724

When using local notifications as a fallback, we still need to respect the users DND status.


SystemKeeper:
I just noticed that this PR inadvertently fixes a problem introduced in https://github.com/nextcloud/talk-ios/pull/859/commits/95dbc10432397b81bff828e2f66a74de863194f8. The moved check should've checked for `serverNotification.notificationType == kNCNotificationTypeChat` instead of `serverNotification.notificationType != kNCNotificationTypeChat`.

Ivansss (approved):

PR #975, opened 2022-11-07
Add completion blocks to external signaling requests

Ivansss:
- [x] Add completion block to joinRoom request
- [x] Reduce websocket timeout

### DIFF
--- a/NextcloudTalk.xcodeproj/project.pbxproj
+++ b/NextcloudTalk.xcodeproj/project.pbxproj
@@ -178,6 +178,7 @@
 		2C62B02424C1BDCF007E460A /* NCAppBranding.m in Sources */ = {isa = PBXBuildFile; fileRef = 2C2A788D2359CC8800EEB797 /* NCAppBranding.m */; };
 		2C62B02624C1BDCF007E460A /* NCUtils.m in Sources */ = {isa = PBXBuildFile; fileRef = 2C5E957B227097E0009CA9BE /* NCUtils.m */; };
 		2C62B02E24C1BDD7007E460A /* PlaceholderView.m in Sources */ = {isa = PBXBuildFile; fileRef = 2CC7158B20B8394A0045C789 /* PlaceholderView.m */; };
+		2C69323D2923ECAA00017AD2 /* WSMessage.m in Sources */ = {isa = PBXBuildFile; fileRef = 2C69323C2923ECAA00017AD2 /* WSMessage.m */; };
 		2C6E74462386D33200AE396C /* ReplyMessageView.m in Sources */ = {isa = PBXBuildFile; fileRef = 2C6E74452386D33200AE396C /* ReplyMessageView.m */; };
 		2C6E7449238C1A0800AE396C /* QuotedMessageView.m in Sources */ = {isa = PBXBuildFile; fileRef = 2C6E7448238C1A0800AE396C /* QuotedMessageView.m */; };
 		2C7381562106136000CDB8DB /* NCChatTitleView.m in Sources */ = {isa = PBXBuildFile; fileRef = 2C7381552106136000CDB8DB /* NCChatTitleView.m */; };
@@ -558,6 +559,8 @@
 		2C62AFB724C1A4E6007E460A /* ShareViewController.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = ShareViewController.h; sourceTree = "<group>"; };
 		2C62AFB824C1A4E6007E460A /* ShareViewController.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = ShareViewController.m; sourceTree = "<group>"; };
 		2C67905128D35BEB00762744 /* sl */ = {isa = PBXFileReference; lastKnownFileType = text.plist.stringsdict; name = sl; path = sl.lproj/Localizable.stringsdict; sourceTree = "<group>"; };
+		2C69323B2923ECAA00017AD2 /* WSMessage.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = WSMessage.h; sourceTree = "<group>"; };
+		2C69323C2923ECAA00017AD2 /* WSMessage.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = WSMessage.m; sourceTree = "<group>"; };
 		2C6E74442386D33200AE396C /* ReplyMessageView.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = ReplyMessageView.h; sourceTree = "<group>"; };
 		2C6E74452386D33200AE396C /* ReplyMessageView.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = ReplyMessageView.m; sourceTree = "<group>"; };
 		2C6E7447238C1A0800AE396C /* QuotedMessageView.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = QuotedMessageView.h; sourceTree = "<group>"; };
@@ -1295,6 +1298,8 @@
 				2C2E64241F3462AF00D39CE8 /* NCSignalingMessage.m */,
 				2CD5F3222142781A006B71BF /* NCExternalSignalingController.h */,
 				2CD5F3232142781A006B71BF /* NCExternalSignalingController.m */,
+				2C69323B2923ECAA00017AD2 /* WSMessage.h */,
+				2C69323C2923ECAA00017AD2 /* WSMessage.m */,
 			);
 			name = WebRTC;
 			sourceTree = "<group>";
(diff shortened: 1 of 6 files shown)


SystemKeeper:
I tried to reproduce some errors I was seeing after trying this PR. Most notably I multiple times got into a state, were I was unable to join the external signaling controller -> because of this PR unable to failed to join the room.

Reading the documentation of `didCloseWithCode`:
> Tells the delegate that the WebSocket task received a close frame from the server endpoint, optionally including a close code and reason from the server.

(https://developer.apple.com/documentation/foundation/nsurlsessionwebsocketdelegate/3181188-urlsession), this does not cover all cases of a connection loss.

On the other hand `didCompleteWithError`:
> The only errors your delegate receives through the error parameter are client-side errors, such as being unable to resolve the hostname or connect to the host. To check for server-side errors, inspect the response property of the task parameter received by this callback.

(https://developer.apple.com/documentation/foundation/nsurlsessiontaskdelegate/1411610-urlsession)

Adding `didCompleteWithError` show the following behavior:

1. Error is raised in `receiveMessageWithCompletionHandler` as well as in `didCompleteWithError`:
```
2022-11-11 18:44:43.1200 WebSocket receiveMessageWithCompletionHandler error Error Domain=kNWErrorDomainPOSIX Code=53 "Software caused connection abort" UserInfo={NSDescription=Software caused connection abort} wss://.../standalone-signaling/spreed
2022-11-11 18:44:43.1270 URLSession didCompleteWithError: Error Domain=kNWErrorDomainPOSIX Code=53 "Software caused connection abort" UserInfo={NSDescription=Software caused connection abort, _NSURLErrorRelatedURLSessionTaskErrorKey=(
    "LocalWebSocketTask <D7C6493B-121B-4EE3-96F1-F1C3466F3B54>.<1>"
), _NSURLErrorFailingURLSessionTaskErrorKey=LocalWebSocketTask <D7C6493B-121B-4EE3-96F1-F1C3466F3B54>.<1>} wss://.../standalone-signaling/spreed
```

As this is handled by the `receiveMessageWithCompletionHandler` method, we don't see a problem in real world usage.

2. Error is only raised in `didCompleteWithError`
```
2022-11-11 22:34:41.4330  Joined room nq48hsid in NC successfully.
2022-11-11 22:34:41.4360  Trying to join room nq48hsid in external signaling server...
2022-11-11 22:34:41.4390  Failed joining room nq48hsid in external signaling server.
2022-11-11 22:34:41.4420  Error joining room, retrying. 0
2022-11-11 22:34:41.4430  URLSession didCompleteWithError: Error Domain=kNWErrorDomainPOSIX Code=57 "Socket is not connected" UserInfo={NSDescription=Socket is not connected, _NSURLErrorRelatedURLSessionTaskErrorKey=(
    "LocalWebSocketTask <761A909B-29C6-482C-9E8C-2E5C062BBEEA>.<1>"
), _NSURLErrorFailingURLSessionTaskErrorKey=LocalWebSocketTask <761A909B-29C6-482C-9E8C-2E5C062BBEEA>.<1>} wss://.../standalone-signaling/spreed
2022-11-11 22:34:41.7240  Joined room nq48hsid in NC successfully.
2022-11-11 22:34:41.7350  Trying to join room nq48hsid in external signaling server...
2022-11-11 22:34:41.7430  Failed joining room nq48hsid in external signaling server.
2022-11-11 22:34:41.7500  Error joining room, retrying. 1
2022-11-11 22:34:41.9920  Joined room nq48hsid in NC successfully.
2022-11-11 22:34:42.0020  Trying to join room nq48hsid in external signaling server...
2022-11-11 22:34:42.0090  Failed joining room nq48hsid in external signaling server.
```

This fails after some trying and the join fails. As multiple tries to join happen, we can be sure that the completion block was actually called. This happened in the completionHandler of the websocket sendmessage method. 

So at this point I think there are some ways to further improve reliability:
* Implement `didCompleteWithError` delegate and in case of an error, reconnect. This might lead to multiple calls to reconnect (from `receiveMessageWithCompletionHandler` and from `didCompleteWithError`. I did notice, that the timeout does not occur after 15s as specified, looks rather like 60s. We should look into this in a follow up and maybe add a timer ourselves.
* In the websockets sendMessage completion handler, if an error occurred, reconnect. The completion handler will always produce a error, if the socket was closed and we try to send a message (I guess we could omit this, if `didCompleteWithError` is there and actually starts a reconnect, but needs to be tested). 
* What might make sense (that might depend on what the decision is above), increase the number of join tries by 1 (so try it 4x) and if we reach the last try, do a reconnect in the external signaling server.
* Maybe we should think again about having a timeout for the completion blocks ...? So if we didn't get a response after 15s, execute the block with an error or something.

Also I did notice that the join attempts are only resetted after a successful join. I'm not sure at this point if this is intended or an oversight, because if I try to join a room again (which failed before), I now only try to join once. By design?

If we are in the error state, we can observe the following
```
(lldb) po _webSocket.state
NSURLSessionTaskStateCompleted

(lldb) po _webSocket.error
Error Domain=kNWErrorDomainPOSIX Code=60 "Operation timed out" UserInfo={NSDescription=Operation timed out}
```

If not, it looks like this
```
(lldb) po _webSocket.state
NSURLSessionTaskStateRunning

(lldb) po _webSocket.error
 nil
```

So before sending a message, we also could check the state/error property of the web socket and initiate a reconnect if needed.

How can this be reproduced?
For me it is reproducible if I start the app while being connected to wifi (so the external signaling server actually resolves to an internal IP) and then disabling wifi to fall back to LTE (so the resolved IP is not accessible anymore). After this I'm unable to join a room again and no reconnection happens.

SystemKeeper:
So I did run a version of talk-ios with the following implemented:
```objc
- (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didCompleteWithError:(NSError *)error
{
    [NCUtils log:[NSString stringWithFormat:@"MM: URLSession didCompleteWithError: %@ %@", error.description, _serverUrl]];
    if (error && task == _webSocket) {
        [NCUtils log:[NSString stringWithFormat:@"MM: WebSocket didCompleteWithError - same obj %@", _serverUrl]];
        [self reconnect];
    }
}
```

It worked better as far as I can tell, but I still ended up in a situation were joining a room was not possible anymore.
While in that state, I did some debugging and it turned out, that the state of the websocket was still running (with error being nil), but apparently I never received a `hello`-response back, so we never did move to `connected` state. Therefore the room join messages piled up in `_pendingMessages`:
```
 (lldb) po _pendingMessages
<__NSArrayM 0x2800e6160>(
<WSMessage: 0x280875660>,
<WSMessage: 0x280875ec0>,
<WSMessage: 0x280ba34c0>,
<WSMessage: 0x2808b5a40>,
<WSMessage: 0x28081f4c0>,
<WSMessage: 0x280825020>,
<WSMessage: 0x28082e820>,
<WSMessage: 0x28084aa20>,
<WSMessage: 0x28082a1e0>,
<WSMessage: 0x280fa0b80>,
<WSMessage: 0x2808b43c0>,
<WSMessage: 0x280b90840>
)
```

Without a chance to ever get send. As we never moved to `connected`, we never tried to send an actual message through the websocket, so the completion handler was never called and we never tried to join a room again. And because we were still in running state of the socket, we never received another `didOpenWithProtocol` call, were we would send a hello request.

So, I think we should ensure that:
* Implement the `didCompleteWithError` delegate (looks like we can then remove the one in the receive message method..)
* We have completion block when sending a `hello` request
* Ensure that completion blocks are called at some point
* Force a reconnect if we try too often to join the room

For the guarantee, I can imagine doing something like this:

```objc
static NSTimeInterval kSendMessageTimeoutInterval = 15;

@interface WSMessage : NSObject
@property NSDictionary *message;
@property SendMessageCompletionBlock completionBlock;
@end

@implementation WSMessage

- (instancetype)initWithMessage:(NSDictionary *)message
{
    self = [super init];
    if (self) {
        self.message = message;
    }
    return self;
}

- (instancetype)initWithMessage:(NSDictionary *)message withCompletionBlock:(SendMessageCompletionBlock)block
{
    self = [self initWithMessage:message];
    if (self) {
        self.completionBlock = block;
        [NSTimer scheduledTimerWithTimeInterval:kSendMessageTimeoutInterval target:self selector:@selector(executeCompletionBlockFailure) userInfo:nil repeats:NO];
    }
    return self;
}

- (void)executeCompletionBlock:(NSError *)error
{
    if (self.completionBlock) {
        self.completionBlock(error);
        self.completionBlock = nil;
    }
}

- (void)executeCompletionBlockFailure
{
    NSError *error = [NSError errorWithDomain:NSCocoaErrorDomain code:0 userInfo:nil];
    [self executeCompletionBlock:error];
}

- (void)executeCompletionBlockSuccess
{
    [self executeCompletionBlock:nil];
}

@end
```

(Maybe store the NSTimer and invalidate it if the block was already called)

SystemKeeper:
Ah, because of pending messages the timer should only be started when we actually tried to send a message, so we can't start it in the init. But I think the idea is clear 😉 
Also `createWebSocketMessage` could be moved to `WSMessage` then...

SystemKeeper:
I pushed some fixes to this PR. Most notably to is the one to make sure that the timeoutTimer is scheduled and invalidated on the main thread in all cases. Also that makes sure that the completion blocks are called on the main thread in all cases.
The last commit is probably the one we need to discuss, but it made sense for me.

SystemKeeper (approved):

PR #978, opened 2022-11-09
Use shared deck card in case there is no reference data available

### LINKED ISSUE #934: Fallback when permission for a deck card is denied
## Steps to reproduce
1. User 1 shares a deck card to a group conversation
2. Users which are part of the group conversation are not necessarily allowed to access said deck card

### Expected behaviour
The shared information should be displayed

### Actual behaviour
Due to using the reference provider, only the link to the deck card will be shown, but the rich object information are discarded.

SystemKeeper:
If a deck card is shared to a conversation, this does not imply that users of that conversation have access to that deck card. If that's the case, the reference provider will return no data and instead of a deck card, the URL is shown. With this PR we now use the information provided by the shared deck card to show at least the title of the card.

Fixes #934

### DIFF
--- a/NextcloudTalk/ChatMessageTableViewCell.m
+++ b/NextcloudTalk/ChatMessageTableViewCell.m
@@ -380,7 +380,15 @@
         }
 
         [message getReferenceDataWithCompletionBlock:^(NCChatMessage *message, NSDictionary *referenceData, NSString *url) {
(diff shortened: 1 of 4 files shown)


SystemKeeper:
Fixes #934

Ivansss (approved):

PR #987, opened 2022-11-10
Update JDStatusBarNotification to v2.0 APIs

Ivansss:


SystemKeeper (approved):
Tested, works and looks nice 😉 

In my opinion would also a good option to replace the toast lib..

PR #1016, opened 2022-12-11
Fix disconnected and joined state

SystemKeeper:
* When we leave a room, correctly reset `hasJoinedRoom` property to `false`
* Currently when we reconnect, we don't set `disconnected` to `true`. This can lead to a situation, where the websocket is `nil`, but won't be connected, because of the missing property.


SystemKeeper:
Also we need to make sure, that the external signaling server is not disconnected while we are in a call. When the app moves to the background, it still starts the timer and disconnects the websocket, although we are in call.

Ivansss (approved):

Issue #252, opened 2019-08-24
iOS Talk app seems to ignore HTTP


Gunnar-GO:
I understand, that the use of microphone and webcam underlies the need of HTTPS (according to the notification @ webview). But why the app simply throws the message "Nextcloud server not found".

May it could write down that only https is supported, better would it be to work with both protocols (HTTP and HTTPS). 

btw. the request to write the protocol "https://..." is uneeded if it will internal cut off and never used. to simplify for users it could be shorted (easier to write, understand etc. for "just"-users)

Android version i couldn't test, but other responses and tests @forum seem to show, that it runs there. (they descibe that the ios app throws the message not found, the android app works flawless with the server)

SystemKeeper:
Just tested the latest version with http:// instead of https:// and it’s working fine.

SystemKeeper:
To add some additional information here:
Apple requires apps to be ATS compliant, so essential you are required to use HTTPS instead of HTTP. This is also true for Nextcloud Talk on iOS (as noted in [Prerequisites](https://github.com/nextcloud/talk-ios#prerequisites)).
ATS however does not apply to some special cases (see [apples documentation](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html#//apple_ref/doc/uid/TP40009251-SW54)), for example:
- Internet protocol (IP) addresses
- Unqualified host names

So it is possible to connect to `http://192.168.1.1`, but it is not possible to connect to `http://nextcloud.mydomain.com`

There are no plans to change the ATS requirement.

> btw. the request to write the protocol "https://..." is uneeded if it will internal cut off and never used. to simplify for users it could be shorted (easier to write, understand etc. for "just"-users)

Indeed it looks like this is needed, but it will be added automatically if you don't enter it, see here:

https://github.com/nextcloud/talk-ios/blob/3595e0975fba7c2d2d97bfcb980747bcfd1fcdd5/NextcloudTalk/LoginViewController.m#L108-L110

PR #1021, opened 2022-12-16
Share accounts with files ios app

Ivansss:


NextcloudTalk/LoginViewController.m
```diff
@@ -219,6 +244,70 @@ - (void)presentAuthenticationViewWithServerURL:(NSString *)serverURL withUser:(N
...
+
+    self.importButton.hidden = !_importedFilesAccount.count;
+}
+
+- (void)presentImportedAccountsSelector
+{
+    NSMutableArray *importedAccounts = [NSMutableArray new];
+    for (NKDataAccountFile *filesAccount in _importedFilesAccount) {
+        DetailedOption *option = [[DetailedOption alloc] init];
+        option.title = (!filesAccount.alias || [filesAccount.alias isEqualToString:@""]) ? filesAccount.user : filesAccount.alias;
```
SystemKeeper, line 284: Not sure about this one:
```
(lldb) po filesAccount.user
marcel.mueller@nextcloud.com

(lldb) po filesAccount.alias
Marcel Müller
```

Because then the user is prepopulated with `Marcel Müller`, which doesn't make sense. Also this does break the Auth-Controller, it's spinning forever..
`DetailOption` has an `identifier` property, maybe always set that to `fileAccount.user` and use it for the login process?

NextcloudTalk/LoginViewController.m
```diff
@@ -219,6 +244,70 @@ - (void)presentAuthenticationViewWithServerURL:(NSString *)serverURL withUser:(N
...
+    [self presentViewController:accountSelectorNC animated:YES completion:nil];
+}
+
+#pragma mark - DetailedOptionSelector delegate
+
+- (void)detailedOptionsSelector:(DetailedOptionsSelectorTableViewController *)viewController didSelectOptionWithIdentifier:(DetailedOption *)option
+{
+    [self dismissViewControllerAnimated:YES completion:^{
+        [self presentAuthenticationViewWithServerURL:option.subtitle withUser:option.title
+        ];
```
SystemKeeper:
```suggestion
        [self presentAuthenticationViewWithServerURL:option.subtitle withUser:option.title];
```

SystemKeeper: Use `option.identifier` here.

NextcloudTalk/NCAppBranding.m
```diff
@@ -39,6 +39,7 @@ @implementation NCAppBranding
 NSString * const copyright = @"© 2022 Nextcloud GmbH";
 NSString * const bundleIdentifier = @"com.nextcloud.Talk";
 NSString * const groupIdentifier = @"group.com.nextcloud.Talk";
+NSString * const appsGroupIdentifier = @"group.com.nextcloud.apps";
```
SystemKeeper, line 42: Branding-related? Or should the import be disabled on branded clients?

NextcloudTalk/LoginViewController.m
```diff
@@ -219,6 +244,70 @@ - (void)presentAuthenticationViewWithServerURL:(NSString *)serverURL withUser:(N
...
+    NSString *path = [[[[NSFileManager defaultManager] containerURLForSecurityApplicationGroupIdentifier:appsGroupIdentifier] URLByAppendingPathComponent:kNextcloudFilesDatabaseFolder] path];
+    NSURL *accountsFileURL = [[NSURL fileURLWithPath:path] URLByAppendingPathComponent:kTalkAccountsFileName];
+    NSArray *filesAccounts = [[NKCommon shared] readDataAccountFileAt:accountsFileURL];
+    NSArray *talkAccounts = [[NCDatabaseManager sharedInstance] allAccounts];
+
+    _importedFilesAccount = [NSMutableArray new];
+    for (NKDataAccountFile *filesAccount in filesAccounts) {
+        BOOL accountIncluded = NO;
+        for (TalkAccount *talkAccount in talkAccounts) {
+            if ([talkAccount.server isEqualToString:filesAccount.url] && [talkAccount.user isEqualToString:filesAccount.user]) {
```
SystemKeeper: This will compare case sensitive. Should be done case insensitive here to be sure?

NextcloudTalk/NCSettingsController.m
```diff
@@ -107,6 +109,7 @@ - (id)init
         _externalSignalingControllers = [NSMutableDictionary new];
         
         [self configureDatabase];
+        [self createAccountsFile];
```
SystemKeeper, line 112: When removing an account, the file needs to be updates as well.

NextcloudTalk/NCSettingsController.m
```diff
@@ -162,6 +165,48 @@ - (void)setActiveAccountWithAccountId:(NSString *)accountId
...
+
+    // Create accounts data
+    NSMutableArray *accounts = [NSMutableArray new];
+    for (TalkAccount *account in [[NCDatabaseManager sharedInstance] allAccounts]) {
+        NSString *avatarPath = [self copyUserAvatarInPath:path forAccount:account];
+        NKDataAccountFile *accountFileData = [[NKDataAccountFile alloc] initWithUrl:account.server user:account.user alias:account.userDisplayName avatar:avatarPath];
+        [accounts addObject:accountFileData];
+    }
+
+    if (!accounts.count) {
```
SystemKeeper, line 189: Not sure, this would mean that we never remove/empty a file, right? If there are no accounts, the file should be deleted?
Just tested, I removed all accounts from Talk, but still can't import all accounts to files, because the account file is not updated when removing an account and after restarting the app, this condition prevents updating/emptying as well.

SystemKeeper (changes requested):
Do we need to introduce an option to disable importing for branded clients? Currently (as it looks like it to me) it would be possible to import from nextcloud files app to branded talk app or vice versa.

NextcloudTalk/DetailedOptionsSelectorTableViewController.m
```diff
@@ -87,7 +99,16 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
 {
     DetailedOption *option = [_options objectAtIndex:indexPath.row];
     UITableViewCell *cell = [[UITableViewCell alloc] initWithStyle:UITableViewCellStyleSubtitle reuseIdentifier:@"DetailOptionIdentifier"];
-    
+
+    if (_type == DetailedOptionsSelectorTypeAccounts) {
+        AccountTableViewCell *accountCell = [tableView dequeueReusableCellWithIdentifier:@"AccountTableViewCellIdentifier" forIndexPath:indexPath];
+        accountCell.accountNameLabel.text = option.title;
+        accountCell.accountServerLabel.text = [option.subtitle stringByReplacingOccurrencesOfString:@"https://" withString:@""];
```
SystemKeeper: Should be case sensitive and should also cover `http://` ?

Ivansss: Case insensitive check might be needed, but no need to check for "http" since it is not possible to connect to a NC server using "http" due to ATS.

SystemKeeper: It's possible when using a IP instead of FQDN: https://github.com/nextcloud/talk-ios/issues/252#issuecomment-1158975761
But I agree, it's probably an edge case...

NextcloudTalk/LoginViewController.m
```diff
@@ -219,6 +244,72 @@ - (void)presentAuthenticationViewWithServerURL:(NSString *)serverURL withUser:(N
...
+    [self presentViewController:accountSelectorNC animated:YES completion:nil];
+}
+
+#pragma mark - DetailedOptionSelector delegate
+
+- (void)detailedOptionsSelector:(DetailedOptionsSelectorTableViewController *)viewController didSelectOptionWithIdentifier:(DetailedOption *)option
+{
+    [self dismissViewControllerAnimated:YES completion:^{
+        [self presentAuthenticationViewWithServerURL:option.subtitle withUser:option.identifier
+        ];
```
SystemKeeper:
```suggestion
        [self presentAuthenticationViewWithServerURL:option.subtitle withUser:option.identifier];
```

SystemKeeper (approved):
Prepouplating the username does not work for me. Not sure if this is a talk-ios issue, though...

PR #1022, opened 2022-12-18
Joining room fixes

SystemKeeper:
Summary:

* When we receive `no_such_session` error, we now gracefully recover by sending the hello message again without a resumeId (no need to reconnect, as the socket is still alive)
* In the external signaling controller we now distinguish between application errors and socket errors. When we receive an application error (like `no_such_room`), there's no need to instantly reconnect the whole websocket.
* The join process in `NCRoomsManager` got refactored, the main point being that we now ignore completion blocks from the external signaling controller, when we're not joining the room anymore or we're joining with a different sessionId. This previously lead to an increase in the join attempts when multiple join processes were running in parallel and therefore ended in `Could not join the conversation` message.

For details, this happened in one of my tests with the current master:
```
2022-12-17 9:07:40.8050 DidReceiveMessage: {"type":"welcome","welcome":{"version":"1.0.0-1~ubuntu20.04","features":["audio-video-permissions","hello-v2","incall-all","mcu","simulcast","transient-data","update-sdp","virtual-sessions","welcome"]}}
2022-12-17 9:07:40.8130 Sending: {"type":"hello","hello":{"version":"1.0","resumeid":"..."},"id":"1"}
2022-12-17 9:07:40.9600 DidReceiveMessage: {"id":"1","type":"hello","hello":{"version":"1.0","sessionid":"...","resumeid":"...","userid":"marcelm","server":{"version":"1.0.0-1~ubuntu20.04","features":["audio-video-permissions","hello-v2","incall-all","mcu","simulcast","transient-data","update-sdp","welcome"]}}}
2022-12-17 9:07:40.9730 Sending: {"type":"room","id":"2","room":{"roomid":"2738090896","sessionid":"Zd31zcVomeN..."}}
2022-12-17 9:07:40.9860 Sending: {"type":"room","id":"3","room":{"roomid":"2738090896","sessionid":"U+FF4ArJdde..."}}
2022-12-17 9:07:40.9980 Sending: {"type":"room","id":"4","room":{"roomid":"2738090896","sessionid":"\/X3SXE56Z0..."}}
2022-12-17 9:07:41.0030 DidReceiveMessage: {"type":"event","event":{"target":"roomlist","type":"update","update":{"roomid":"2738090896","properties":{"name":"Private Unterhaltung","type":1,"lobby-state":0,"lobby-timer":null,"read-only":0,"listable":0,"active-since":null,"sip-enabled":0,"participant-list":"refresh"}}}}
2022-12-17 9:07:41.0100 DidReceiveMessage: {"type":"event","event":{"target":"roomlist","type":"update","update":{"roomid":"2738090896","properties":{"name":"Private Unterhaltung","type":1,"lobby-state":0,"lobby-timer":null,"read-only":0,"listable":0,"active-since":null,"sip-enabled":0,"participant-list":"refresh"}}}}
2022-12-17 9:07:41.0140 DidReceiveMessage: {"type":"event","event":{"target":"roomlist","type":"update","update":{"roomid":"2738090896","properties":{"name":"Private Unterhaltung","type":1,"lobby-state":0,"lobby-timer":null,"read-only":0,"listable":0,"active-since":null,"sip-enabled":0,"participant-list":"refresh"}}}}
2022-12-17 9:07:41.1010 DidReceiveMessage: {"id":"2","type":"error","error":{"code":"no_such_room","message":"The user is not invited to this room."}}
```

So because the websocket wasn't connected, there were multiple join attempts for the same room, but with different sessionIds. Obviously two of them must fail, because the sessionId was invalid at this point. The error `no_such_room` then triggered a reconnect and execution of all message completion blocks with an error, leading to the aforementioned `Could not join..` error. We now handle application errors without an immediate reconnect and also ignore completion blocks of previous join attempts, which are already superseded by another join attempt.

### DIFF
--- a/NextcloudTalk/WSMessage.m
+++ b/NextcloudTalk/WSMessage.m
@@ -138,7 +138,6 @@ static NSTimeInterval kSendMessageTimeoutInterval = 15;
     }
 
     NSLog(@"Sending: %@", self.webSocketMessage);
-    [NCUtils trace:[NSString stringWithFormat:@"Sending: %@", self.webSocketMessage]];
     NSURLSessionWebSocketMessage *message = [[NSURLSessionWebSocketMessage alloc] initWithString:self.webSocketMessage];
     [webSocketTask sendMessage:message completionHandler:^(NSError * _Nullable error) {
         if (error && self.completionBlock) {
--- a/NextcloudTalk/WSMessage.m
+++ b/NextcloudTalk/WSMessage.m
@@ -70,6 +70,15 @@ static NSTimeInterval kSendMessageTimeoutInterval = 15;
     return NO;
 }
 
+- (BOOL)isJoinMessage
+{
+    if ([[_message objectForKey:@"type"] isEqualToString:@"room"]) {
+        return YES;
+    }
+
+    return NO;
+}
+
 - (void)setMessageTimeout
 {
     // NSTimer uses the runloop of the current thread. Only the main thread guarantees a runloop, so make sure we dispatch it to main!
--- a/NextcloudTalk/WSMessage.m
+++ b/NextcloudTalk/WSMessage.m
@@ -21,6 +21,7 @@
  */
 
 #import "WSMessage.h"
+#import "NCUtils.h"
 
 static NSTimeInterval kSendMessageTimeoutInterval = 15;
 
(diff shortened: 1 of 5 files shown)


NextcloudTalk/WSMessage.m
```diff
@@ -134,10 +129,11 @@ - (void)sendMessageWithWebSocket:(NSURLSessionWebSocketTask *)webSocketTask
     }
 
     NSLog(@"Sending: %@", self.webSocketMessage);
+    [NCUtils trace:[NSString stringWithFormat:@"Sending: %@", self.webSocketMessage]];
```
Ivansss: Do you want to add trace() to this PR or was it just for testing purposes? :)

Ivansss (changes requested):

NextcloudTalk/WSMessage.m
SystemKeeper: Oops, I added it locally, but wasn't meant for the PR 😅

Ivansss (approved):

PR #1029, opened 2023-01-03
Fix checkUnreadMessagesVisibility

Ivansss:
dispatch_async checkUnreadMessagesVisibility so indexPathForMessage is called on main thread.

This PR fixes crash reported in Xcode.


SystemKeeper (approved):
Makes sense 👍

PR #1042, opened 2023-01-18
Do not create publisher peer connection when no media is going to be sent

Ivansss:
Fixes problem connecting to the call without neither audio nor video publishing permissions.


SystemKeeper (approved):
Tested with and without permission and worked 👍

PR #1047, opened 2023-01-19
Use batch update for collection view

### LINKED ISSUE #528: Don’t redraw the call screen immediately
1. Have a call with 40 people
2. Call is over and people start leaving
3. After each user the call screen is recalculated and redrawn jumping videos all the time.

I think it would were good to delay this for e.g 100-200ms and then redraw.
Would still be lots of moving but at least not look as odd

SystemKeeper:
Fixes #528 

Delay updating the collection view a bit and do it as a batch update. If there's no participant in the call the participant is directly added.

### DIFF
--- a/NextcloudTalk/CallViewController.m
+++ b/NextcloudTalk/CallViewController.m
@@ -54,6 +54,18 @@ typedef NS_ENUM(NSInteger, CallState) {
     CallStateInCall
 };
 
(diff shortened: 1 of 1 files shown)


SystemKeeper:
Still not sure if 200ms is too low for this 🤔

Ivansss (approved):

PR #1048, opened 2023-01-19
Use peer dispayName for generating background color

Ivansss:


SystemKeeper (approved):
Works with my test cases as well 👍

PR #1060, opened 2023-01-31
🖖 Breakout rooms

### LINKED ISSUE #981: 🖖 Breakout rooms
Meta ticket at https://github.com/nextcloud/spreed/issues/8337

- [ ] TBD

Ivansss:
Fix #981

### DIFF
--- a/NextcloudTalk/NCExternalSignalingController.m
+++ b/NextcloudTalk/NCExternalSignalingController.m
@@ -500,7 +500,7 @@ static NSTimeInterval kWebSocketTimeoutInterval = 15;
 
 - (void)processSwitchToMessageEvent:(NSDictionary *)messageDict
 {
-    NSString *roomToken = [messageDict objectForKey:@"roomId"];
+    NSString *roomToken = [messageDict objectForKey:@"roomid"];
     if (roomToken.length > 0) {
         [self.delegate externalSignalingController:self shouldSwitchToCall:roomToken];
     } else {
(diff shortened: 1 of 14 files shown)


NextcloudTalk/NCExternalSignalingController.m
```diff
@@ -496,6 +498,16 @@ - (void)processRoomMessageEvent:(NSDictionary *)messageDict
     }
 }
 
+- (void)processSwitchToMessageEvent:(NSDictionary *)messageDict
+{
+    NSString *roomToken = [messageDict objectForKey:@"roomId"];
```
SystemKeeper: https://github.com/nextcloud/spreed/pull/8477/files#diff-7ed631e9f04a0445d27e26f716dee988bd363d56cbfa4453ed74bd3605433f9aR296

Looks like this should be "roomid"?

SystemKeeper:
Some comments:
* Switching is working very smoothly - I like it!
* I had a crash when switching:
  ```
  Assertion failed: (_internal->figCaptureSession == NULL), function -[AVCaptureVideoPreviewLayer attachToFigCaptureSession:]_block_invoke, file AVCaptureVideoPreviewLayer.m, line 1201.
  dyld4 config: DYLD_LIBRARY_PATH=/usr/lib/system/introspection DYLD_INSERT_LIBRARIES=/Developer/usr/lib/libBacktraceRecording.dylib:/Developer/usr/lib/libMainThreadChecker.dylib:/Developer/Library/PrivateFrameworks/DTDDISupport.framework/libViewDebuggerSupport.dylib
   ```
   This only happened once and I was not able to reproduce it again. 
* When switching the old session is not ended? I switched between Room1 and Room2. So initially I joined in Room1 and then switched to Room2, but my last frame was still visible in Room1 and the status on the cell changed to "Failed to connect to...". Sometimes I wasn't able to establish a connection again after that, but that might be a problem of my test env.
* Sometimes the chat is not shown again after switching. Problem is, that there's still a '_chatViewController` sometimes, because the switch was too fast. Using `dispatch_after` looks like the most reliable to me in `callControllerDidJoinCall` (I'd go for 0.5, since the animation is 0.3).

SystemKeeper (approved):

PR #1073, opened 2023-02-09
Calculate the minimum width based on the target aspect ratio

SystemKeeper:
The minimum width of a cell was a fixed value based on a 1.5 aspect ratio. Since we have a target aspect ratio of 1.0 for portrait mode, this lead to weird layouts. Now the minimum width is calculated on the minimum height and the target aspect ratio. Also the spacing is now taken into account.

Before:

After:


Ivansss (approved):
👌

PR #1089, opened 2023-02-22
Fix screensharing

SystemKeeper:
This PR does 2 things:
* Prevent the video transceiver being stopped when it's a screen sharing peer and we're in audio-only mode.
* Revert the dispatch timing changes done in https://github.com/nextcloud/talk-ios/pull/1081 while keeping the correct queue.


SystemKeeper:
I think the first one is not needed actually: 
https://github.com/nextcloud/talk-ios/blob/a8150b35d4db03ad76b161adf4bbaa6667617264/NextcloudTalk/NCCallController.m#L717

Looks like it was masked in my tested by the second issue. @Ivansss ?

SystemKeeper:
Removed first commit and force-pushed

Ivansss (approved):

Issue #992, opened 2022-11-18
Received messages have a mildly rough animation


e-caste:
## Steps to reproduce
1. Have a chat open
2. Receive a message in the chat
3. Observe the messages slightly jump

### Expected behaviour
Smoother animation. I have observed rough/twitchy/jumpy animations in these situations:
- receiving a message with the chat open
- loading new messages / updating the chat after opening a chat (either sent from the current account from another device or received from other participants) - in this case the chat jumps down after having loaded all messages at the same time
- sometimes when opening a chat, there is no animation but the app glitches from the chat list view to the chat view directly

### Actual behaviour
The chat has a very small twitch, which can be annoying when receiving multiple messages.

## Device information

**Device:** iPhone X

**iOS version:** 16.1.1

**Talk version:** 15.1.1

## Server information

**Nextcloud version:** (see admin overview page: `/index.php/settings/admin/overview`) 25.0.1

**Talk version:** (see apps admin page: `/index.php/settings/apps`) 15.0.1

**Custom Signaling server configured:** yes/no and version (see talk admin settings: `/index.php/settings/admin/talk#signaling_server`) no

**Custom TURN server configured:** yes/no (see talk admin settings: `/index.php/settings/admin/talk#turn_server`) yes

**Custom STUN server configured:** yes/no (see talk admin settings: `/index.php/settings/admin/talk#stun_server`) yes

### Server log (data/nextcloud.log)
<details>

```
Insert your server log here
```
</details>

SystemKeeper:
Hey @e-caste, there were some improvements in the latest TestFlight builds about this. Maybe you could take a look if it's working now better for you?

e-caste:
@SystemKeeper Hi, sorry for the delay in the response. I have installed version 16.0.0 from the App Store which I believe includes these changes but I haven't noticed any difference, e.g. when I open a chat with new messages they appear all at once without animations.

SystemKeeper:
Thanks for the info! Maybe I'm misunderstand here, but if you open a chat with unread messages, then yes, there's no animation to expect there.  At least this is how it is at the moment. Might take a look if that could be improved, but I think it will not work smoothly with the unread marker..

e-caste:
@SystemKeeper I see. Then in that case:
- the missing animation when entering a chat appears to be solved
[...]

PR #1099, opened 2023-03-01
Fix chat scroll position when loading multiples blocks

Ivansss:
The unread messages indicator was misplaced when loading multiple blocks.
With this PR we force the chat view to keep current scroll position.
Also, the animation when adding new messages seems smoother now.


SystemKeeper (approved):
Tested and works 👍 
Might help with #992 maybe?

PR #1104, opened 2023-03-02
Use VLCKit to show WebM videos

SystemKeeper:
It's not perfect, especially since VLCKit has quite some bugs (most of them are already solved in the latest alpha for version 4). So for now it's used for webm files so we can view call recordings.


SystemKeeper:
- [x] Fix device rotation
- [x] Add close button
- [x] Prefer light style statusbar

Ivansss (approved):

Issue #1102, opened 2023-03-02
Add backgroundTask to share recording


SystemKeeper:

SystemKeeper:
Done in #1105

PR #1105, opened 2023-03-03
Add background Task to recording notification handling

SystemKeeper:
When we share a recording or dismiss a recording notification while the app is in the background (for example stop recording on Phone 1 but try to share from Phone 2) there is a high chance of the request being cancelled by the OS before it completed. So we need to make sure we keep the app alive until the request finished (as it's done when sending a message).


SystemKeeper:
Fixes #1102

Ivansss (approved):

PR #1106, opened 2023-03-03
Check skip lobby permission

SystemKeeper:
How to test:

```
1 give phone user advanced permission "ignore lobby"
2. join chat on phone
3. enable lobby in web
4. user can see the chat and can write
5. go out and rejoin
6. user can see the chat but not write
```

### DIFF
--- a/NextcloudTalk/NCChatViewController.m
+++ b/NextcloudTalk/NCChatViewController.m
@@ -2722,7 +2722,13 @@ NSString * const NCChatViewControllerTalkToUserNotification = @"NCChatViewContro
 
 - (BOOL)shouldPresentLobbyView
 {
-    if ([[NCDatabaseManager sharedInstance] serverHasTalkCapability:kCapabilityChatPermission] && (_room.permissions & NCPermissionCanIgnoreLobby) != 0) {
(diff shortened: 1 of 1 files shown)


NextcloudTalk/NCChatViewController.m
```diff
@@ -2722,6 +2722,10 @@ - (void)updateRoomInformation
 
 - (BOOL)shouldPresentLobbyView
 {
+    if ([[NCDatabaseManager sharedInstance] serverHasTalkCapability:kCapabilityChatPermission] && (_room.permissions & NCPermissionCanIgnoreLobby) != 0) {
```
Ivansss: We should check for `conversation-permissions` capabilty instead of `chat-permission` capabilty.
Because that's when `Can ignore lobby` premission was introduced.

Since `conversation-permissions` capability was not added in Talk 13 release, you can check it using this workaround:
https://github.com/nextcloud/talk-ios/blob/master/NextcloudTalk/CallViewController.m#L260-L264

Ivansss (changes requested):

Ivansss (approved):

PR #1111, opened 2023-03-06
Improve lobby

### LINKED ISSUE #1107: Improve lobby view
There are still 2 visual things we should improve:
1. When joining with lobby active you can still see the chat for a brief second
2. The text input has not the right height when the lobby gets disabled while already being in a conversation

SystemKeeper:
Fixes #1107


SystemKeeper:
https://github.com/Ivansss/SlackTextViewController/pull/19 is needed to have correct margins and smooth animations.

Ivansss (approved):

PR #1122, opened 2023-03-09
Add sanity check for update participants notification

Ivansss:
Not all room participants events contain `roomid` or `users`, so we need to check those values before adding them to `userinfo` dictionary.
This PR fixes app crash when receiving a room participants event for leaving the call (when a moderator presses "End call for everyone")


SystemKeeper (approved):
Nice, thx!

PR #1125, opened 2023-03-10
Fix user profile image regression

Ivansss:
A regression was introduced in #1113 
`saveProfileImageForAccount` needs to be called with the updated account (the one that contains the userId)


SystemKeeper (approved):
Good catch!

PR #1149, opened 2023-03-17
Add social app icon and use it in contacts menu

Ivansss:


SystemKeeper:
Should we use SF here as well? 
Bildschirm­foto 2023-03-18 um 18 24 10

SystemKeeper (approved):

PR #1161, opened 2023-03-24
Misc automation / optimizations

chenxiao0228:
Changed Cocoapods source to use the CDN registry
Automated the download process of webrtc xcframework
Automated submodule update
Updated README to reflect the change to setup workflow

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,10 +15,7 @@ Nextcloud Talk lowers the barrier for communication and lets your team connect a
 - [CocoaPods](https://cocoapods.org/)
 
 ## Development setup
-run
-```
-pod install && open NextcloudTalk.xcworkspace
-```
+After cloning this repository, you can use `pod install` to install all dependencies. After that, open the project with open `NextcloudTalk.xcworkspace`.
 
 Pull Requests will be checked with [SwiftLint](https://github.com/realm/SwiftLint). We strongly encourage the installation of SwiftLint to detect issues as early as possible.
 
--- a/README.md
+++ b/README.md
@@ -15,15 +15,9 @@ Nextcloud Talk lowers the barrier for communication and lets your team connect a
 - [CocoaPods](https://cocoapods.org/)
 
 ## Development setup
-
-Clone this repository and download the [WebRTC library](#webrtc-library)
-
+run
 ```
-$ pod install
-
-$ git submodule update --init
-
-$ open NextcloudTalk.xcworkspace
+pod install && open NextcloudTalk.xcworkspace
 ```
 
 Pull Requests will be checked with [SwiftLint](https://github.com/realm/SwiftLint). We strongly encourage the installation of SwiftLint to detect issues as early as possible.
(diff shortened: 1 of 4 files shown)


download_webrtc.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -euo pipefail
+
+DOWNLOAD_FILE=$(mktemp  -u)
+
+curl -L "https://github.com/nextcloud-releases/talk-clients-webrtc/releases/download/108.5359.0/WebRTC.xcframework.zip" -o "$DOWNLOAD_FILE"
```
SystemKeeper: Can we move the versioning to a variable?

README.md
```diff
@@ -15,15 +15,9 @@ Nextcloud Talk lowers the barrier for communication and lets your team connect a
 - [CocoaPods](https://cocoapods.org/)
 
 ## Development setup
-
-Clone this repository and download the [WebRTC library](#webrtc-library)
-
+run
```
SystemKeeper: We would still need to clone the repository, so I think it makes sense to state 

> After cloning this repository, you can use `pod install` to install all dependencies. After that, open the project with `open NextcloudTalk.xcworkspace`.

SystemKeeper (commented):
Hello,

thank you very much for your contribution!
I noted a few things on the files. In general:
* I really like the idea of using the pods to install the dependencies 👍 
* Do we need to commit the *xcscheme file? Maybe we should add that to the ignore list
* While I get the idea of using a CDN, nearly everything around nextcloud is based on GitHub, so I don't think we should start using a CDN here.

chenxiao0228:
Thanks @SystemKeeper  for your review!

> * I really like the idea of using the pods to install the dependencies 👍
> * Do we need to commit the *xcscheme file? Maybe we should add that to the ignore list

I noticed there is no default scheme for the app target when I did a clean checkout.
Since it's checked in under xcshareddata, everyone will get it by default.

> * While I get the idea of using a CDN, nearly everything around nextcloud is based on GitHub, so I don't think we should start using a CDN here.

The CDN is the equivalent of the main podspec repo and it is recommended after CP 1.9 or something. You'd get the exact same results as the podspec repo without paying for the overhead of cloning / pulling that gigantic git repo.

SystemKeeper:
> The CDN is the equivalent of the main podspec repo and it is recommended after CP 1.9 or something.

Thanks for letting me know. For reference, here's the official blog post about it: https://blog.cocoapods.org/CocoaPods-1.8.0-beta/

> I noticed there is no default scheme for the app target when I did a clean checkout.
Since it's checked in under xcshareddata, everyone will get it by default.

Thanks again, need to take a close look there 👍

Ivansss (approved):
Thanks a lot for your contribution @chenxiao0228 !
Really nice work :)

SystemKeeper:
Thank you very much @chenxiao0228! I fixed one small typo in the README (I thought I made a comment, but forget obviously).
Feel free to reach out if you would like to contribute in the future here :-)

PR #1179, opened 2023-04-06
Do not show destructive actions in room info view while in a call

Ivansss:
Destructive actions were shown when opening room info view from call view's title view.


SystemKeeper (approved):
Tested and works :)

PR #1183, opened 2023-04-06
Use systemImageName for stop recording

Ivansss:


SystemKeeper (approved):
Oops, did not notice before...

PR #1239, opened 2023-05-22
Reaction shortcut

SystemKeeper:
image


SystemKeeper:
Last commit improves the animation of grouped cells. Before the full cell (with avatar) was visible directly after tapping on the cell. Now only the grouped part is visible and the avatar part will be shown when the context menu is visible as well.
Landscape mode on regular iPhones can still be improved a bit, but that needs some work, so would do it in a follow-up.

Ivansss (approved):

PR #1283, opened 2023-06-14
Add UI test for login and conversation creation

SystemKeeper:


SystemKeeper:
* Added UI tests for login and conversation creation
* Splitted instance creation to parallelise work
* Added test matrix for stable27 and master
* SwiftLint failure due to garm
* Unit tests can be done in a follow up

ci-create-docker-server.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+# Based on the script from Milen Pivchev in nextcloud/ios repository
+
+# This script installs docker and starts a nextcloud test instance
+
+echo "Setup docker..."
+
+brew install docker
+colima start
```
mpivchev: Is this faster than using the 3rd party github action?

NextcloudTalkUITests/NextcloudTalkUITests.swift
```diff
@@ -0,0 +1,161 @@
...
+
+import XCTest
+
+final class NextcloudTalkUITests: XCTestCase {
+
+    let timeoutLong: Double = 60
+    let timeoutShort: Double = 15
+    let server = "http://localhost:8080"
+    let username = "admin"
+    let password = "admin"
```
mpivchev, line 30: You decided to not use env-vars? Isn't it better to have one source for these values that can be changed on the fly? What if I make my server in a different port, or use different credentials, I would have to change all tests to use them.

NextcloudTalkUITests/NextcloudTalkUITests.swift
```diff
@@ -0,0 +1,161 @@
...
+    func launchAndLogin() -> XCUIApplication {
+        let app = XCUIApplication()
+        app.launchArguments += ["-AppleLanguages", "(en-US)"]
+        app.launchArguments += ["-AppleLocale", "\"en-US\""]
+        app.launch()
+
+        let accountSwitcherButton = app.buttons["Nextcloud Talk"]
+
+        // Wait shortly until the app is fully started
+        _ = accountSwitcherButton.waitForExistence(timeout: timeoutShort)
```
mpivchev, line 62: You can do `XCTAssert` to avoid the empty assignment:

```suggestion
        XCTAssert(accountSwitcherButton.waitForExistence(timeout: timeoutShort))
```

.github/workflows/uitests.yml
```diff
@@ -0,0 +1,69 @@
...
+      run: |
+        pod install
+
+    - name: Wait for server
+      run: |
+        source ./ci-wait-for-server.sh
+
+    - name: Install talk
+      run: |
+        source ./ci-install-talk.sh
```
mpivchev, line 64: It would be great if the scripts that we use in all 3 projects have the same name. This will make things easier for someone who switches.

mpivchev (changes requested):
Not sure if the PR is fully complete, but if it's set as ready for review I will assume so.

Some more things are needed:
- Integration tests
- Snapshot tests
- Update to README to include instructions for testing

.github/workflows/uitests.yml
```diff
@@ -0,0 +1,69 @@
...
+      TEST_BRANCH: ${{ matrix.test-branches }}
+      CONTAINER_NAME: nextcloud_test
+      SERVER_PORT: 8080
+
+    steps:
+    - name: Checkout app
+      uses: actions/checkout@v3
+      with:
+        submodules: true
+
```
mpivchev: Aren't we cloning the branch here and then also cloning it in `ci-install-talk.sh`?

ci-create-docker-server.sh
SystemKeeper: Yes, because the 3rd party action always runs a update of homebrew which is not needed here.

NextcloudTalkUITests/NextcloudTalkUITests.swift
SystemKeeper: For these tests, yes. Currently only UI tests are included here, so there's no need to have an app token right now as it is generated in the login process. And since this test is done from the user perspective, I don't see the need to have it in a generated file (username and password are hardcoded in the image anyway).

NextcloudTalkUITests/NextcloudTalkUITests.swift
SystemKeeper: Nope, that would mean the test will fail if the button does not exist. Here it is meant to have a way to detect if there's already an account set up or not (so login is needed or not)

.github/workflows/uitests.yml
SystemKeeper: I thought about this as well. But I think the requirements are different for the projects and there might be more project-specific files and also steps being added in the future (for example you probably don't need to install talk currently). And looking into the future the contents might differ as well, so from my POV this shouldn't be a requirement and will only cause troubles in the long run when a certain file is expected to do a certain thing which it doesn't do.

.github/workflows/uitests.yml
SystemKeeper: No, this one (`actions/checkout`) checks out the talk-ios repository. The one in `ci-install-talk.sh` checks out the spreed (so the server part of talk) repository to install the app on the instance.

SystemKeeper:
> Not sure if the PR is fully complete, but if it's set as ready for review I will assume so.
> 
> Some more things are needed:
> 
> * Integration tests
> * Snapshot tests
> * Update to README to include instructions for testing

Well, the PR is complete for what it tried to achieve ;) It was a first step for us to get some tests up and running, it's not meant to provide a full test suite.

But I agree for the last point, there should be a note in the README.

SystemKeeper:
README updated and instructions added. Moved the docker installation to a workflow step

.github/workflows/uitests.yml
mpivchev: Fair enough 👍

NextcloudTalkUITests/NextcloudTalkUITests.swift
mpivchev: There are only UI tests now, but they still need a docker image and a server, and IMO it's good to add envars now for future proofing. But ultimately the decision is up to you.

NextcloudTalkUITests/NextcloudTalkUITests.swift
```diff
@@ -0,0 +1,161 @@
...
+        waitForExpectations(timeout: timeoutLong, handler: nil)
+    }
+
+    func waitForHittable(object: Any?) {
+        let enabledPredicate = NSPredicate(format: "hittable == true")
+        expectation(for: enabledPredicate, evaluatedWith: object, handler: nil)
+        waitForExpectations(timeout: timeoutLong, handler: nil)
+    }
+
+    func launchAndLogin() -> XCUIApplication {
```
mpivchev, line 53: I think a more descriptive format for test functions is needed. For example:
```
test_functionThatIsTested_withCircumstances_shouldExpectation()
```

so in this case it can become something like:

```
test_logIn_withProperParams_shouldLogInSuccessfully()
```

ci-create-docker-server.sh
mpivchev: Hmm, not sure it's a good idea to expect different states in a test. IMO it's better to reset the state and test only one kind of state. In the Files integration and ui tests I reset all accounts before testing log in.

NextcloudTalkUITests/NextcloudTalkUITests.swift
SystemKeeper: But this function is not a test which runs on its own, that’s why it’s not prefixed with „test“?

ci-create-docker-server.sh
SystemKeeper: I think this was meant for the other conversation, right? :)
In general I totally agree. Each test should be independent, but since the runners on macOS are ridiculously slow, I went this way for now. 
If we find the time, we can rework some stuff in talk-ios and add some Startup parameters which allow us separate these tests.

NextcloudTalkUITests/NextcloudTalkUITests.swift
```diff
@@ -0,0 +1,161 @@
...
+
+        // Open profile menu
+        profileButton.tap()
+
+        // At this point we should be logged in, so check if username and server is displayed somewhere
+        XCTAssert(app.staticTexts[server].waitForExistence(timeout: timeoutShort))
+        XCTAssert(app.staticTexts[username].waitForExistence(timeout: timeoutShort))
+    }
+
+    func testCreateConversation() {
```
mpivchev, line 132: What about this 😄

NextcloudTalkUITests/NextcloudTalkUITests.swift
```diff
@@ -0,0 +1,161 @@
...
+
+        // When the account switcher gets enabled, we have atleast 1 account in the app and are online
+        XCTAssert(accountSwitcherButton.waitForExistence(timeout: timeoutLong))
+        waitForEnabled(object: accountSwitcherButton)
+
+        return app
+    }
+
+    // Tests are done in alphabetical order, so we want to always test login first
+    func test_AAAA_Login() {
```
mpivchev, line 117: This also doesn't follow the convention.

Ivansss (approved):

mpivchev:
@Ivansss there were some things i mentioned that were not resolved in the PR, but I probably should've requested changes.

PR #1289, opened 2023-06-27
Filter conversation list for unread / mentioned conversations

### LINKED ISSUE #1264: Filter conversation list for unread / mentioned conversations only
* Overview issue in https://github.com/nextcloud/spreed/issues/9668

Ivansss:
IMG_E0002

Fix #1264

### DIFF
--- a/NextcloudTalk/RoomsTableViewController.m
+++ b/NextcloudTalk/RoomsTableViewController.m
@@ -50,10 +50,17 @@
 
 typedef void (^FetchRoomsCompletionBlock)(BOOL success);
 
+typedef enum RoomsFilter {
+    kRoomsFilterAll = 0,
+    kRoomsFilterUnread,
+    kRoomsFilterMentioned
+} RoomsFilter;
+
 @interface RoomsTableViewController () <CCCertificateDelegate, UISearchBarDelegate, UISearchControllerDelegate, UISearchResultsUpdating>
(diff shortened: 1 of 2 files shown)


SystemKeeper:
Working good in general 👍 
Found some issues with filtering and searching.

Case 1:
* Filter for "Unread"
* Search for a term where no results are returned
* Cancel search

Expected: All rooms visible again
Actual: Only unread rooms visible (even though "All" is selected)



Case 2:
* Filter for unread
* Search for a term where no results are returned
* Remove the search term

Expected: Only unread rooms should be visible
Actual:


Both cases seem to have something to do with the 30s timer to update the rooms or atleast the 30s timer influences something here as well.

Ivansss:
Nice catches! ;)

Case 1 fixed in 0b45173ee64b17ad306b21f998e4aa99194821be
Case 2 fixed in 96befff72bf3cfa266760c4321539b15d914ae49

SystemKeeper (approved):
Works :-) 

I just noticed that we now have the case that we call `self.tableView reloadData` twice when `refreshRoomList` is called and we don't have a search string. Once it's done inside of `filterRooms` and once in `refreshRoomList`. Not sure if this will be a problem in the real world or if iOS might even optimize the two calls, but we should keep that in mind and optimize that in the long run.

Issue #1290, opened 2023-07-01
PlaceholderView is overlapped by searchbar in landscape mode


SystemKeeper:
Ref #1287 

![image](https://github.com/nextcloud/talk-ios/assets/1580193/db6deef5-4390-4d3a-ba7b-ca798fd9aa24)

PR #1301, opened 2023-07-12
Adjust placeholder view constraints to safe area

Ivansss:


SystemKeeper:
image

🤔

Ivansss:
I couldn't find an easy fix for this problem, not as trivial as I thought.
This PR fixes the positioning of the activity indicator.
This issue of the screenshot above is tracked here https://github.com/nextcloud/talk-ios/issues/1290 and we can try to fix it in a follow-up PR.

SystemKeeper:
> This issue of the screenshot above is tracked here #1290 and we can try to fix it in a follow-up PR.

Sorry, I though that should be the fix for exactly that…

SystemKeeper (approved):
Now that I know what I should test, this works as expected ;)

PR #1303, opened 2023-07-14
Use singular `File.exist?` alias in Podfile

fwcd:
This fixes the `pod install` when using Ruby 3.2+ since `File.exists?` has been removed.

See here: https://stackoverflow.com/questions/14351272/undefined-method-exists-for-fileclass-nomethoderror


Ivansss (approved):

Ivansss:
Thanks a lot for your contribution @fwcd :)

PR #1328, opened 2023-08-14
Add markdown property to NCChatMessage

### LINKED ISSUE #1299: Markdown support
* Overview ticket: https://github.com/nextcloud/spreed/issues/1027
* API discussion: https://github.com/nextcloud/spreed/issues/9950

We plan to role out the following support for markdown:

# non-optional
- [x] bold
- [x] italic

# optional
Optional means it is just shown with the markdown syntax

- [x] strike-through
- [x] list
- [ ] todo list with checkboxes
- [x] code block using 3 `
- [x] inline code using `
- [x] quote
- [x] headlines
- [ ] dividers `---`

SystemKeeper:
Fixes #1299 

Render markdown only if we receive the `markdown` property for the message.


nickvergessen (approved):
🤷🏼 

Didn't look bad per se

PR #1329, opened 2023-08-14
Test UI tests on M1 runners

SystemKeeper:


SystemKeeper:
Well, no M1 runners available right now…

PR #1330, opened 2023-08-14
Fix appStateHasChanged in SettingsTableViewController

### LINKED ISSUE #1326: Failing to get User Status in Settings
## Steps to reproduce
1. Run App on simulator (Could not test if it happens on device aswell)
2. Open Settings
3. Watching "Fetching User Status"

### Expected behaviour
Fetch the user status after some seconds

### Actual behaviour
Some times (not consistently) the first time the settings are opened, after launch the "getActiveUserStatus()" method fails, or at least does not return the actual status. Should maybe be retried if error appears.

## Device information

**Device:** Any Simulator

**iOS version:** 16.2

**Talk version:** 17.1

https://github.com/nextcloud/talk-ios/blob/68df3b5b8d35d1667ded5aae1be7a9bdcc71c151/NextcloudTalk/SettingsTableViewController.swift#L237-L245

SystemKeeper:
Fixes #1326 

We can't directly convert a `typedef enum` from ObjC to the bridged Swift enum. Therefore the `if let` construct fails and we never update the UI after the app state was changed.

### DIFF
--- a/NextcloudTalk/SettingsTableViewController.swift
+++ b/NextcloudTalk/SettingsTableViewController.swift
@@ -248,7 +248,8 @@ class SettingsTableViewController: UITableViewController, UITextFieldDelegate {
 
     @objc func appStateHasChanged(notification: NSNotification) {
         let appState = notification.userInfo?["appState"]
-        if let appState = appState as? AppState {
+        if let rawAppState = appState as? UInt32 {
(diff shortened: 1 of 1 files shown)


Black-Fox-2022 (approved):
Solves the problem of not loading

PR #1355, opened 2023-09-04
Update webrtc 116.5845.0

Ivansss:


SystemKeeper (approved):
Tested and works 👍 

Maybe we should think about adding the lib also via SPM: https://developer.apple.com/documentation/xcode/distributing-binary-frameworks-as-swift-packages

SystemKeeper:
Judging from https://webrtc.googlesource.com/src.git/+/44bc8e96ed88005fec89a1cc479e291fea30d1b3/sdk/objc/components/renderer/metal/RTCMTLI420Renderer.mm#122 the fix for the 1x1 crashes we were seeing and patching ourselves should now be part of the 116 version. Can you confirm that from the sources you used to build the lib? We could then go ahead and remove our patch from the lib repo...

Ivansss:
Yes, the fix is on the source files we used to build the lib.
Should we remove (and  don't apply) the patch for the following version?
https://github.com/nextcloud-releases/talk-clients-webrtc/blob/main/ios/patches/RTCMTLVideoView.patch

Or do you think we should build a new version (e.g. 116.5845.1) without the patch?

SystemKeeper:
> Or do you think we should build a new version (e.g. 116.5845.1) without the patch?

I would suggest dropping our patch in the next version upgrade. Our patch fixes the same issue, just in a different way/position. So for 116 it shouldn't matter one way or another.

Issue #1317, opened 2023-07-30
The app doesn't connect to the external signaling server after the server has been restarted


e-caste:
## Steps to reproduce
1. Update then bring up the Nextcloud server (in my case, with `docker compose up -d`)
2. Use Talk iOS: it very quickly connects to the signaling server in every chat
3. Back up the server (in my case, with `docker compose down`, then back up, then `docker compose up -d`)

### Expected behaviour
Talk iOS can connect to the signaling server without issue.

### Actual behaviour
Talk iOS can't connect to the signaling server anymore, which I notice from the disabled call buttons:
![image](https://github.com/nextcloud/talk-ios/assets/48513706/03755444-ee50-49b2-b871-322fd8be655c)
The effects of this are:
- I can't perform any calls in any chat
- I can't see the typing indicator
- the app takes 10-15 seconds to send a message (which is unfeasibly long when you're outside - the sending of the message can be triggered more quickly by tapping back at the top left and then re-joining the chat, which is horrible UX)

## Device information

**Device:** iPhone X

**iOS version:** 16.5.1 (c)

**Talk version:** 17.0.0

## Server information

**Nextcloud version:** (see admin overview page: `/index.php/settings/admin/overview`) 27.0.1

**Talk version:** (see apps admin page: `/index.php/settings/apps`) 17.0.3

**Custom Signaling server configured:** yes/no and version (see talk admin settings: `/index.php/settings/admin/talk#signaling_server`) yes, app says version Unknown (when it could connect it showed a commit number), I have configured strukturag/nextcloud-spreed-signaling:1.1.3 which is the latest one

**Custom TURN server configured:** yes/no (see talk admin settings: `/index.php/settings/admin/talk#turn_server`) yes

**Custom STUN server configured:** yes/no (see talk admin settings: `/index.php/settings/admin/talk#stun_server`) yes

### Server log (data/nextcloud.log)
<details>

```
Insert your server log here
```
</details>

SystemKeeper:
Hey e-caste,

just to state the obvious, do backups only outside of working hours 😉 
Jokes aside... Some things to note:

* Sending a message does not rely on the external signaling server, so when sending a message takes 10-15s after a restart, I would assume it's because the NC instance is not responding fast enough. Maybe you could verify the requests on the server in that case? You could also verify on web and take a look at the requests. That could also be an issue for the signaling server, because it needs to request capabilities from the NC instance.
[...]

PR #1358, opened 2023-09-10
Support "already_joined" error message

SystemKeeper:
This PR does 2 things:

* Supports `already_joined` error (introduced at https://github.com/strukturag/nextcloud-spreed-signaling/pull/547)
  * With this we get a error message in case we try to join a room, that we already joined in the external signaling server. In case we are aware that we joined this room before, we treat this error as a successful join attempt. In any other case, we treat it as a error and try to reconnect to the server.
* Does a `forceReconnect` when we our message to join a room times out
  * Currently when `joinRoom` fails (e.g. we run into a timeout), we do a reconnect and try to resume our session. But in this case, we would try to join the room again, which would fail again as we don't receive a confirmation/error about that duplicated join and after 3 tries, the whole process will fail. So for the next join attempt, we don't try to resume our session, but do a fresh start.

Related to #1317, but keeping open until we know if there's something wrong with 1.1.3.

### DIFF
--- a/NextcloudTalk/NCExternalSignalingController.m
+++ b/NextcloudTalk/NCExternalSignalingController.m
@@ -181,6 +181,7 @@ NSString * const NCExternalSignalingControllerDidReceiveStoppedTypingNotificatio
 {
     dispatch_async(dispatch_get_main_queue(), ^{
         self->_resumeId = nil;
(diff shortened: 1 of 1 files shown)


Ivansss (approved):
Awesome 👌

PR #1369, opened 2023-09-21
Fix missing screensharing indicators

Ivansss:
Fixes problem with screensharing indicators not shown in users that are sharing their screen.


SystemKeeper (approved):
Tested and works - nice!
Failing test related to docker workaround, fix at #1371

PR #1378, opened 2023-09-29
📝 Note to self

### LINKED ISSUE #300: Support "Note to self"
Similar implementation to the changelog room.

"bot" and "notes" are the magical words.

Avatar is based on the same background as the app logo +notes icon for the foreground: https://material.io/resources/icons/?search=notes&icon=notes&style=baseline logo.

Ivansss:
Fix #300

### DIFF
--- a/NextcloudTalk/BaseChatViewController.swift
+++ b/NextcloudTalk/BaseChatViewController.swift
@@ -926,6 +926,35 @@ import QuickLook
         self.presentWithNavigation(shareViewController, animated: true)
     }
 
+    func didPressNoteToSelf(for message: NCChatMessage) {
+        let activeAccount = NCDatabaseManager.sharedInstance().activeAccount()
+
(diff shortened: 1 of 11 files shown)


SystemKeeper (changes requested):
Noted some differences to web:
* In web it's allowed to change the name/description and avatar of the room, on iOS it's forbidden
* "Note to self" context menu item should not be available when inside the "Note to self" room
* Web allows to clear the messages
* Web allows to delete the room (we have "Leave conversation", which makes the room disappear because the only participant left, but guess that's not the intention). Web sends DELETE to `ocs/v2.php/apps/spreed/api/v4/room/...`

SystemKeeper:
Also in Web the Start call button is completely hidden, should we hide it as well instead of disabling?

SystemKeeper (approved):
Working nicely now! 👍

PR #1382, opened 2023-10-12
Open room links inside the app

Ivansss:


SystemKeeper:
I haven’t tested this yet, but just looking at the code, would it make sense to do the check case insensitive?

SystemKeeper (approved):
Nice 👍

PR #1384, opened 2023-10-14
Dependencies

RomanPodymov:
Hello.
Thank you for talk-ios.
Updated dependencies.

### DIFF
--- a/NextcloudTalk/NCAPIController.m
+++ b/NextcloudTalk/NCAPIController.m
@@ -241,7 +241,7 @@ NSInteger const kReceivedChatMessagesLimit = 100;
                                  @"search" : phoneNumbers};
     
     NCAPISessionManager *apiSessionManager = [_apiSessionManagers objectForKey:account.accountId];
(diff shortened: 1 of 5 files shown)


Ivansss:
Thanks a lot for your contribution @RomanPodymov :) 
We are thinking about moving to Alamofire since AFNetworking is deprecated, because of that we are wondering if there's a particular reason for bumping the library from v3 to v4? If not, we would prefer to stay at version 3 for the time being.
We would be happy to add the rest of the changes from your PR.

RomanPodymov:
> Thanks a lot for your contribution @RomanPodymov :) We are thinking about moving to Alamofire since AFNetworking is deprecated, because of that we are wondering if there's a particular reason for bumping the library from v3 to v4? If not, we would prefer to stay at version 3 for the time being. We would be happy to add the rest of the changes from your PR.

Hello @Ivansss 
Reverted AFNetworking.

Ivansss (approved):

PR #1401, opened 2023-10-26
fix(i18n): Fixed typo

rakekniven:


SystemKeeper:
Thanks @rakekniven !
I’ll take a look later to make CI happy :)

SystemKeeper (approved):
How many people does it take to add a character 🙈

rakekniven:
> How many people does it take to add a character 🙈 

Fixing such typos isn't fun at all. 
I do not blame anyone. We are humans and we make errors. 
I am hoping to have an automatic spellchecker at GitHub soon. 
Having that at Transifex is too late in the whole chain.

SystemKeeper:
I wrote this comment, because my fix commit was not correct (looks like on the latest MacOS there's an encoding change), so Ivan committed again to fix the CI.

> Fixing such typos isn't fun at all.

Is there anything we can help with that in regards to talk-ios? For us it's usually quite easy, because we generate the localizable files with a script (if you're using a mac yourself, see the `generate-localizable-strings-file.sh` script) and only need to adjust the source.

PR #1403, opened 2023-10-26
Setup spell checker for localizable file

SystemKeeper:
Add spell check to localizable CI. Added it there to always check with the latest localizable file (otherwise the run would first fail for not correctly updating the localizable file and then for spell checking).

Test of failure: https://github.com/nextcloud/talk-ios/actions/runs/6657060665/job/18090981017?pr=1403

### DIFF
--- a/.github/workflows/localizable.yml
+++ b/.github/workflows/localizable.yml
@@ -38,9 +38,18 @@ jobs:
       run: |
         git diff
 
+    - name: Install dependencies
+      run: |
+        pip3 install pyspelling
+
+    - name: Spell check
+      run: |
+        python3 -m pyspelling
+
     - name: Verify Changed files
       uses: tj-actions/verify-changed-files@v16
       id: verify-changed-files
+      if: always()
       with:
         fail-if-changed: true
         files: |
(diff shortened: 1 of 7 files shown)


SystemKeeper:
CC @rakekniven

rakekniven:
Hello @SystemKeeper ,

really appreciate you picked up my comment and added something right away.

Does that mean something similar has to be added for each repo?
If so, is there any mechanism to have that globally for all NC apps incl. server? (dreaming in the clouds 😄  )

Ivansss (approved):

SystemKeeper:
> If so, is there any mechanism to have that globally for all NC apps incl. server? (dreaming in the clouds 😄 )

Indeed there are templates for CI. The main issue I see is adjusting a spell checker to the corresponding repository. In iOS it's quite easy, there's a single file that's generated and committed and is the only one that needs to be checked. AFAIK on a javascript project these strings are extracted from the source files directly by the transifex job, so not sure how this would be set up there. Another issue might be resources on CI, on busy days there's already quite some waiting.

PR #1404, opened 2023-10-27
Update referencing messages when a message was updated

SystemKeeper:
Before:

image

After:

Bildschirmfoto 2023-10-27 um 11 26 15


SystemKeeper:
Just had an idea, maybe it would be more performant if we iterate through the visibleIndexPaths and check if any visible message has a corresponding parent…?

Ivansss (approved):

PR #1332, opened 2023-08-16
UserProfilStatus to SwiftUI

Black-Fox-2022:
The three views, connected to the Status are now written in SwiftUI. Two of them look exactly the same as before, but the third now also has the standart iOS default design language.

# Main View (same Design)

Bildschirmfoto 2023-08-16 um 12 17 48

# Status View (same Design)

Bildschirmfoto 2023-08-16 um 12 18 13

# Status Message View (before)

Bildschirmfoto 2023-08-16 um 12 12 38

# Status Message View (after)

Bildschirmfoto 2023-08-16 um 12 17 58

PR #1417, opened 2023-11-10
User status SwiftUI fixes

Ivansss:
Follow up of #1332 

- [x] Remove disclosure indicator in settings view for user status option since we are not presenting it as a modal
- [x] Update settings view when dismissing user status view.
- [x] Do not use default navigation style (SplitViewStyle on iPads).
- [x] Use same background colors for buttons as for tableview in UserStatusMessage view.
- [x] Add some padding to buttons in UserStatusMessage view.

### DIFF
--- a/NextcloudTalk/UserStatusSwiftUIView.swift
+++ b/NextcloudTalk/UserStatusSwiftUIView.swift
@@ -23,7 +23,7 @@ import UIKit
 import SwiftUI
 import SwiftUIIntrospect
 
-protocol UserStatusViewDelegate{
+protocol UserStatusViewDelegate: AnyObject {
     func userStatusViewDidDisappear()
 }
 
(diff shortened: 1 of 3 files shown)


Black-Fox-2022 (approved):

NextcloudTalk/UserStatusSwiftUIView.swift
```diff
@@ -23,6 +23,10 @@ import UIKit
 import SwiftUI
 import SwiftUIIntrospect
 
+protocol UserStatusViewDelegate{
```
SystemKeeper:
```suggestion
protocol UserStatusViewDelegate: AnyObject {
```

NextcloudTalk/UserStatusSwiftUIView.swift
```diff
@@ -33,6 +37,8 @@ struct UserStatusSwiftUIView: View {
         _userStatus = State(initialValue: userStatus)
     }
 
+    var delegate: UserStatusViewDelegate?
```
SystemKeeper: Should this be a weak var?

SystemKeeper:
IMG_A7DA37693269-1

Happens on iOS 15, don't see it on Simulator with iOS 17...

PR #1434, opened 2023-11-28
Use default background color in RoomAvatarInfoTableView

Ivansss:
Before:
IMG_E0028

After:
IMG_E0027


SystemKeeper (approved):
Wasn't even noticable in dark mode 😅

PR #1435, opened 2023-11-30
Allow to set room description

Ivansss:
IMG_E0029

### DIFF
--- a/NextcloudTalk/NCAPIController.h
+++ b/NextcloudTalk/NCAPIController.h
@@ -41,6 +41,7 @@ typedef void (^GetRoomsCompletionBlock)(NSArray *rooms, NSError *error, NSIntege
 typedef void (^GetRoomCompletionBlock)(NSDictionary *roomDict, NSError *error);
 typedef void (^CreateRoomCompletionBlock)(NSString *token, NSError *error);
 typedef void (^RenameRoomCompletionBlock)(NSError *error);
+typedef void (^SetDescriptionCompletionBlock)(NSError *error);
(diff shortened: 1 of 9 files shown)


SystemKeeper:
image

The inputAccessoryView does look weird when you have an external keyboard attached. What is the reason to have this and not have it when changing the title?

I also noticed (not part of this PR) that we can select the title cell:
image

Description is now editable in the conversation settings: 
image

Ivansss:
> The inputAccessoryView does look weird when you have an external keyboard attached. What is the reason to have this and not have it when changing the title?

The room title is inside a textfield and we use the "Done" button in the keyboard to confirm changes.
Room description is inside a textview so there is no "Done" button in the keyboard (instead there is the new line button), so we need to add a confirmation button somewhere in the UI.

> The inputAccessoryView does look weird when you have an external keyboard attached. What is the reason to have this and not have it when changing the title?

It does look weird I agree 😅 , the reason was explained above :) 

> I also noticed (not part of this PR) that we can select the title cell

That could be fixed, already fixed it for the description but I couldn't select the title.

> Description is now editable in the conversation settings

That was done intentionally, but we can discuss if we really want that or only allow to edit title and description in RoomAvatarInfoTableViewController

SystemKeeper:
> That was done intentionally, but we can discuss if we really want that or only allow to edit title and description in RoomAvatarInfoTableViewController

For me it felt weird because it wasn’t clear how to save it. Probably because I missed the done button 😅

SystemKeeper (approved):

PR #1191, opened 2023-04-17
Conversation avatars (View part)

SystemKeeper:
Ref #979

- [ ] Fix Force-Reloading the own profile avatar

PR #1447, opened 2023-12-09
Refresh room list on account change

SystemKeeper:
Regression from https://github.com/nextcloud/talk-ios/pull/1191

Since we only listen to changes to `NCRoom` we don't immediately refresh the rooom list, when the active account is changed. This is problematic in case the server of the new active account can't be reached. In this case we still show the old room list, while the active account already changed.


SystemKeeper:
Superseded by https://github.com/nextcloud/talk-ios/pull/1459

PR #1449, opened 2023-12-09
Add background processing

SystemKeeper:
While the app refresh is based on the usage of the app, background processing on the other hand should be independent of the usage. Let's try to add background processing additionally for removing already read notifications.


SystemKeeper:
So let's get this in to help with notifications. When we want to background update rooms/messages, we should make sure that processing and refreshing are not run in parallel.

Ivansss (approved):

PR #1477, opened 2023-12-27
Check existence of selectedPredifinedStatus

SystemKeeper:
Fixes a crash seen on TestFlight.

Steps to reproduce:
- Set a custom user status
- Open `UserStatusMessageSwiftUIView` again
- Without changing anything, press `Set status message`
- 💣

Reason is that `selectedPredifinedStatus` is force un-wrapped and it is `nil` in that case.

### DIFF
--- a/NextcloudTalk/UserStatusMessageSwiftUIView.swift
+++ b/NextcloudTalk/UserStatusMessageSwiftUIView.swift
@@ -224,8 +224,8 @@ struct UserStatusMessageSwiftUIView: View {
     }
 
     func setActiveUserStatus() {
-        if !customStatusSelected {
-            NextcloudKit.shared.setCustomMessagePredefined(messageId: selectedPredifinedStatus!.id!, clearAt: selectedClearAt) { _, error in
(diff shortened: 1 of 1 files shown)


Black-Fox-2022 (approved):

SystemKeeper:
Looking at this again, I am wondering if it would be better to disable the save button until something was really changed by the user. Any idea on that @Black-Fox-2022 ? 🤔

Black-Fox-2022:
> Looking at this again, I am wondering if it would be better to disable the save button until something was really changed by the user. Any idea on that @Black-Fox-2022 ? 🤔

Yeah, that’s definitely possible. I can try that next week.

Black-Fox-2022:
> > Looking at this again, I am wondering if it would be better to disable the save button until something was really changed by the user. Any idea on that @Black-Fox-2022 ? 🤔
> 
> Yeah, that’s definitely possible. I can try that next week.

Uploaded a new Pull request with the changes, unfortunatly couldn't use this PR as base missed some file I couldn't figure out why (which wasn't missing in the other branch)

Ivansss (approved):

PR #1478, opened 2023-12-28
Remove face rectangle detection

SystemKeeper:
When we enable background blur, we issue a face rectangle request and a segmentation request. Only the segmentation request is really needed to correctly blur the background, so we can remove the face rectangle request. Might also be worth checking, if this improves performance on old iOS devices.


Ivansss:
> Might also be worth checking, if this improves performance on old iOS devices.

There is a slight improvement of the performance when testing it on an iPhone X :)

Ivansss (approved):

PR #1479, opened 2023-12-28
Check camera permission and request access if needed

SystemKeeper:
The app crashes when camera access is denied and we try to start/join a video call:
- Disable camera access in the iOS settings
- Try to join a video call
- 💣

Also I noticed that since we switched to `NCCameraController` for camera access, we never request access to the camera anymore (I assume this was done by WebRTC before).

This PR does the following:
- Checks if camera access was granted before trying to access the camera
- If camera access is not available, the camera button in the `CallViewController` is disabled
- When starting a call, we check if camera access was requested, if not, we request it.

### DIFF
--- a/NextcloudTalk/CallViewController.m
+++ b/NextcloudTalk/CallViewController.m
@@ -1958,7 +1958,7 @@ typedef void (^UpdateCallParticipantViewCellBlock)(CallParticipantViewCell *cell
 - (void)callController:(NCCallController *)callController userPermissionsChanged:(NSInteger)permissions
 {
     [self setAudioMuteButtonEnabled:(permissions & NCPermissionCanPublishAudio)];
-    [self setVideoDisableButtonEnabled:(permissions & NCPermissionCanPublishVideo)];
+    [self setVideoDisableButtonEnabled:((permissions & NCPermissionCanPublishVideo) && [callController isCameraAccessAvailable])];
 }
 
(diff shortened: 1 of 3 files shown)


Ivansss:
> Also I noticed that since we switched to NCCameraController for camera access, we never request access to the camera anymore (I assume this was done by WebRTC before).

That's not the case in my tests 🤔 it requests camera access even without the changes in this PR.
The crash is reproducible.

SystemKeeper:
> That's not the case in my tests 🤔

Just retested and I am now asked for permission as well 🤔 I am pretty sure that in my tests that did not work, because I had to select "Camera" from sharing to see the permission dialog. Weird.
Shall we remove https://github.com/nextcloud/talk-ios/pull/1479/files#diff-4924e88da89128e4862d29acf861ed8f2193a745b7c350546046e978ae62e462R157-R166 and keep the rest?

Ivansss:
> Shall we remove https://github.com/nextcloud/talk-ios/pull/1479/files#diff-4924e88da89128e4862d29acf861ed8f2193a745b7c350546046e978ae62e462R157-R166 and keep the rest?

Yes :) I think we can remove that part of the code. The rest of the PR makes total sense ;)

SystemKeeper:
> Yes :)

Done 👍

Ivansss (approved):

PR #1489, opened 2024-01-15
Add background task for decreasing unread badge number

Ivansss:
Attempt to fix following app crash:

Screenshot 2024-01-15 at 10 59 20


SystemKeeper:
I am wondering if we should add a background task to the calling function, `removeNotificationsWithNotificationIds` in this case?

Ivansss:
> I am wondering if we should add a background task to the calling function, removeNotificationsWithNotificationIds in this case?

moved :)

SystemKeeper (approved):

PR #1499, opened 2024-01-25
Test Realm binary

SystemKeeper:
Let's see if this is faster on CI ...


SystemKeeper:
Seems to work. Saves minimum of 25% of CI time. We could add the repo to the deps org. Current example see https://github.com/SystemKeeper/realm-swift-binary/blob/main/Package.swift

SystemKeeper:
Included in https://github.com/nextcloud/talk-ios/pull/1497

PR #1525, opened 2024-02-09
Fix screensharing disappearing issue

### LINKED ISSUE #1466: Received screenshare disappears after joining
## Steps to reproduce
1. As admin start a call and share your screen on web
2. As user2 join the call from iOS

### Expected behaviour
After joining the screenshare should be displayed

### Actual behaviour
It looks like the screenshare is shortly visible (you see the close button for a second), but vanishes afterwards. Also the button to show the screenshare on the participant is missing.

https://github.com/nextcloud/talk-ios/assets/1580193/99ccb0a4-e6bd-4643-8bc7-99d6374bce2d

## Device information

**Device:** iPhone XR and iPhone 11 Pro

**iOS version:** 17.2

**Talk version:** master

## Server information

**Nextcloud version:** master

**Talk version:** master

**Custom Signaling server configured:** yes

Ivansss:
How to test:

1. Start a call with user A and start screensharing
2. Join call with user B using the iOS app

Without this PR:
User B joins the call and most of the time the screensharing appears just for a second and then vanishes.

With this PR:
User B joins the call and can see User A screen share.

Fix  #1466

### DIFF
--- a/NextcloudTalk/CallViewController.m
+++ b/NextcloudTalk/CallViewController.m
@@ -2460,9 +2460,6 @@ typedef void (^UpdateCallParticipantViewCellBlock)(CallParticipantViewCell *cell
                 [[[peer getRemoteStream].videoTracks firstObject] removeRenderer:videoRenderer];
             }];
 
-            // Screen renderers
-            [self removeScreensharingOfPeer:peer];
-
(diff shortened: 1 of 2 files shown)


SystemKeeper (approved):
Nice catch 👍

PR #1547, opened 2024-02-21
Add airplay option when speaker button is inside menu

Ivansss:


NextcloudTalk/CallViewController.m
```diff
@@ -926,10 +925,25 @@ - (void)adjustMoreButtonMenu
...
+            speakerActionTitle = NSLocalizedString(@"Audio options", nil);
+        }
+
+        void (^speakerBlock)(UIAction *action) = ^void(UIAction *action) {
             [weakSelf speakerButtonPressed:nil];
-        }];
+        };
+
+        void (^airplayBlock)(UIAction *action) = ^void(UIAction *action) {
+            for (id subview in self->_airplayView.subviews) {
```
SystemKeeper: Pretty sure this will lead to a retain cycle again because we capture a strong reference to self. We need to do something like this:

https://github.com/nextcloud/talk-ios/blob/43cf853b6fe1b1e2aab08d28f7c693247410ecd1/NextcloudTalk/FileMessageTableViewCell.m#L302-L303

SystemKeeper (changes requested):

NextcloudTalk/CallViewController.m
Ivansss: Good catch! :)

SystemKeeper (approved):
Very nice!

PR #1549, opened 2024-02-22
Fix user reconnected grouped system messages

Ivansss:
Before:
User joins the call and then leaves the call.
Both system messages were grouped as "reconnected"

With this PR:
Only when the user leaves the call and then joins again, both system messages are grouped as "reconnected"


SystemKeeper (approved):
Good catch!

PR #1564, opened 2024-03-06
Federation capabilities

SystemKeeper:


SystemKeeper:
Storing the capabilities and checking the proxy-hash is added. Also we reload a chat view now when capabilities for that room changed.

Ivansss (approved):

PR #1581, opened 2024-03-18
Use xcbeautify to reduce the warning noise of xcodebuild

SystemKeeper:
- [ ] Seems `xcbeautify` is already available on the images, so can skip the install step


SystemKeeper:
Added to https://github.com/nextcloud/talk-ios/pull/1579

PR #1587, opened 2024-03-22
Show federated avatars when creating a new room and adding a federated participant

Ivansss:
Now when creating a new room and adding federated participants, its avatars will be shown.


SystemKeeper (approved):
Tested and works 👍

PR #1608, opened 2024-04-02
Retry status page curl command

SystemKeeper:
Seems like sometimes the curl command is too fast to connect to the status page (see https://github.com/nextcloud/talk-ios/actions/runs/8529679303/job/23366327951?pr=1606). So we should retry the command before failing.


SystemKeeper:
Still fails sometimes. We should try to ignore the error code for now and see if our tests still work correctly.

Ivansss (approved):

PR #1622, opened 2024-04-12
Include calls in recents setting

Ivansss:
IMG_E0034 1


SystemKeeper (approved):
Tested and works - awesome!

PR #1629, opened 2024-04-19
Always end call for everyone in one to one conversations

Ivansss:
Based on #1628 

Draft until we implement it in all platforms

### DIFF
--- a/NextcloudTalk/CallViewController.m
+++ b/NextcloudTalk/CallViewController.m
@@ -881,15 +881,26 @@ typedef void (^UpdateCallParticipantViewCellBlock)(CallParticipantViewCell *cell
 
         [self adjustMoreButtonMenu];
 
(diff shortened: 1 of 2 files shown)


SystemKeeper:
Wondering if we should just change the default for 1:1 here. Tap -> End for all, Long press -> Option to just leave the call. I still can see occasions that I would like to leave a call without making the other leave the call as well? 🤔

nickvergessen:
Clarifying expectation:
- One-to-one:
  - Short/Normal press: "End call" (Label is short, but behaviour is "End call for everyone")
  - Long press: "Leave call"
- Group:
  - Short/Normal press: "Leave call"
  - Long press: "End call for everyone"

SystemKeeper (approved):
Tested and works 👍

PR #1634, opened 2024-04-24
Fix NCCallController retain cycle

### LINKED ISSUE #1467: Sharing iOS screen sometimes deadlocks
## Steps to reproduce
1. ?!

Not fully sure yet, sometimes the worker thread locks up and any webrtc related task fails because of this. Needs investigation.

### LINKED ISSUE #1627: screensharing not working when initiated from iphone
## Steps to reproduce
1. make a video call to another participant
2. initiate screenshare from iphone 

### Expected behaviour
initiating screen share on iphone shares screens of iphone to another iphone connected to the call or person connected via a browser. 

### Actual behaviour
it works 10 percent most of the time after i start the broadcast nothing happens , the other person sees only the first initial of the other person. however when initiating screen share from the web browser to web browser or web browser to iphone it works everytime as expected. I have tested this from 2 different iphones with the same result 

## Device information

**Device:** iphone 12 

**iOS version:** 17.4.1

**Talk version:** 18.0.3

nextcloud version 28.0.3

SystemKeeper:
Fixes #1627 
Fixes #1467 

When a call is ended, we remove the observer from the darwin notification center, but since the handler is still stored in the array, we retain the `NCCallController`. Since the starting/stopping of broadcasting is systemwide, all `NCCallControllers` will try to handle it and create publishers. 
An alternative would be to use a `weakself` in the handler itself, but this should be more failsafe

### DIFF
--- a/.github/workflows/uitests.yml
+++ b/.github/workflows/uitests.yml
@@ -9,6 +9,7 @@ on:
       - NextcloudTalkTests/**
       - NotificationServiceExtension/**
       - ShareExtension/**
+      - BroadcastUploadExtension/**
 
     push:
       branches:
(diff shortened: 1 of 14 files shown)


SystemKeeper:
I am unable to reproduce #1467 with this PR anymore. So I guess it had the same root cause.

Ivansss (approved):
Really nice work 👌

PR #1662, opened 2024-05-20
Move realm methods to swift

SystemKeeper:
I first tried to move `NCRoom` completely to Swift, but since we are still using the ObjC version of Realm, some features are not available, e.g. optionals. So moving the properties over itself, should be done when we are also switching to RealmSwift.

For now I moved all* the methods of `NCRoom` and `NCChatMessage` over to Swift. Also using the `NS_ENUM` or `NS_OPTIONS` macros to have better Swift support for our enums/options.

*) Some methods are left in `NCChatMessage` since they access private variables and might conflict otherwise with Realm when making them public. Need some more testing in a follow-up.

Best to review commit by commit.

Also removed some of the non-used properties:
```
- Property 'NCRoom.roomId' has been removed.
- Property 'NCRoom.count' has been removed.
- Property 'NCRoom.lastPing' has been removed.
- Property 'NCRoom.numGuests' has been removed.
- Property 'NCRoom.guestList' has been removed.'
```


Ivansss (approved):
Really nice 👌

PR #1674, opened 2024-06-07
New one to one conversation creation flow

Ivansss:
| Conversation list | 1. Create a new conversation <br> 2. Join open conversation <br> 3. Create one2one conversation  | Create a new conversation  |
|---|---|---|
| IMG_E0049 | IMG_E0046 | IMG_E0048 |

### DIFF
--- a/NextcloudTalk/ContactsSearchResultTableViewContoller.swift
+++ b/NextcloudTalk/ContactsSearchResultTableViewContoller.swift
@@ -1,0 +1,109 @@
+//
+// Copyright (c) 2024 Ivan Sein <ivan@nextcloud.com>
+//
+// Author Ivan Sein <ivan@nextcloud.com>
+//
+// GNU GPL version 3 or any later version
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+
+import UIKit
+
+@objcMembers class ContactsSearchResultTableViewController: UITableViewController {
+
+    var indexes: [String] = []
+    var contacts: [String: [NCUser]] = [:]
+
+    let tableBackgroundView = PlaceholderView()
+
+    override init(style: UITableView.Style) {
+        super.init(style: style)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        self.tableView.register(UINib(nibName: kContactsTableCellNibName, bundle: nil), forCellReuseIdentifier: kContactCellIdentifier)
+
+        tableBackgroundView.setImage(UIImage(named: "contacts-placeholder"))
+        tableBackgroundView.placeholderTextView.text = NSLocalizedString("No results found", comment: "")
+        self.showSearchingUI()
+        self.tableView.backgroundView = tableBackgroundView
+
+    }
+
+    // MARK: - UI
+
+    func showSearchingUI() {
+        tableBackgroundView.placeholderView.isHidden = true
+        tableBackgroundView.loadingView.startAnimating()
+        tableBackgroundView.loadingView.isHidden = false
+    }
+
+    func hideSearchingUI() {
+        tableBackgroundView.loadingView.stopAnimating()
+        tableBackgroundView.loadingView.isHidden = true
+    }
+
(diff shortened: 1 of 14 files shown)


NextcloudTalk/NewRoomTableViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+    case joinOpenRoom
+}
+
+@objcMembers class NewRoomTableViewController: UITableViewController,
+                                               UINavigationControllerDelegate,
+                                               UISearchResultsUpdating {
+
+    var account: TalkAccount
+
+    var indexes: [String] = [""]
```
SystemKeeper, line 35: This feels wrong 🤔 Is it to align the section number to the index?

NextcloudTalk/NewRoomTableViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+        super.viewWillAppear(animated)
+
+        self.navigationItem.hidesSearchBarWhenScrolling = false
+        self.getPossibleContacts()
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        self.navigationController?.navigationBar.titleTextAttributes = [NSAttributedString.Key.foregroundColor: NCAppBranding.themeTextColor()]
```
SystemKeeper: Could use https://github.com/nextcloud/talk-ios/blob/master/NextcloudTalk/NCAppBranding.m#L228 until we have a super class for this.

NextcloudTalk/NewRoomTableViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+            }
+        }
+    }
+
+    // MARK: - Search
+
+    func updateSearchResults(for searchController: UISearchController) {
+        self.resultTableViewController.showSearchingUI()
+        DispatchQueue.main.async {
+            self.searchTimer = Timer.scheduledTimer(timeInterval: 1.0, target: self, selector: #selector(self.searchForContacts), userInfo: nil, repeats: false)
```
SystemKeeper, line 151: I think there's a missing invalidation here. Shouldn't the search timer be invalidated at some point to have a correct debounce?

NextcloudTalk/NewRoomTableViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+        let navController = NCNavigationController(rootViewController: viewController)
+        self.present(navController, animated: true)
+    }
+
+    func createRoomWithContact(_ contact: NCUser) {
+        NCAPIController.sharedInstance().createRoom(forAccount: account, withInvite: contact.userId, ofType: .oneToOne, andName: nil) { room, error in
+            if let token = room?.token, error == nil {
+                self.navigationController?.dismiss(animated: true, completion: {
+                    NotificationCenter.default.post(name: NSNotification.Name.NCSelectedUserForChat, object: self, userInfo: ["token": token])
+                })
```
SystemKeeper: Could also write it like: 

```suggestion
                self.navigationController?.dismiss(animated: true) {
                    NotificationCenter.default.post(name: NSNotification.Name.NCSelectedUserForChat, object: self, userInfo: ["token": token])
                }
```

NextcloudTalk/NewRoomTableViewController.swift
Ivansss: This is basically the section we'll use for the "Create a new conversation" and "Join open conversations" options.
We can take another (more self-explainatory) approach sure ;)

NextcloudTalk/NewRoomTableViewController.swift
Ivansss: Forgot about that function. I agree, we should use it until we create a super class 👍

NextcloudTalk/NewRoomTableViewController.swift
Ivansss: I was pretty sure I added the timer invalidation here :P

NextcloudTalk/NewRoomTableViewController.swift
Ivansss: Fixed!

NextcloudTalk/NewRoomTableViewController.swift
Ivansss: Fixed!

NextcloudTalk/ContactsSearchResultTableViewContoller.swift
```diff
@@ -0,0 +1,109 @@
...
+//
+
+import UIKit
+
+@objcMembers class ContactsSearchResultTableViewController: UITableViewController {
+
+    var indexes: [String] = []
+    var contacts: [String: [NCUser]] = [:]
+
+    let tableBackgroundView = PlaceholderView()
```
SystemKeeper:
```suggestion
    let tableBackgroundView = PlaceholderView(for: .insetGrouped)!
```


https://github.com/nextcloud/talk-ios/assets/1580193/a1a31f82-9e69-4310-b6ff-c07e83ba73ed

SystemKeeper (approved):
One small design nit-pick, otherwise really nice job 👌

NextcloudTalk/NewRoomTableViewController.swift
SystemKeeper: Not really sure how to solve it easily, without counting differently in all other methods. I'd say we keep it for now like this?!

PR #1676, opened 2024-06-10
Correctly handle "too_many_requests" error

SystemKeeper:
Currently we try to resume a session every time when we know a `resumeId`. This is problematic after changes to the signaling server, namely https://github.com/strukturag/nextcloud-spreed-signaling/pull/744/commits/4c807c86e83dbcb2958b9d4da0ad75d0d9417588. 

This PR fixes this by:
* Correctly handling `too_many_requests` error message
* Only try to resume a session, when the disconnect was less than 30s ago

### DIFF
--- a/NextcloudTalk/NCExternalSignalingController.m
+++ b/NextcloudTalk/NCExternalSignalingController.m
@@ -59,6 +59,7 @@ NSString * const NCExternalSignalingControllerDidReceiveStoppedTypingNotificatio
 @property (nonatomic, assign) NSInteger reconnectInterval;
 @property (nonatomic, strong) NSTimer *reconnectTimer;
 @property (nonatomic, assign) BOOL sessionChanged;
+@property (nonatomic, assign) NSInteger disconnectTime;
 
 @end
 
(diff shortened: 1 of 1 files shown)


SystemKeeper:
I am wondering, do we need to set the disconnectTime also in one of the delegates ? But then we don’t know, when the disconnect really happened?! Or only in the close delegate?

Ivansss (approved):

Ivansss:
I guess it makes sense to add it in the close delegate.
It might make sense to also add it `- (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didCompleteWithError:(NSError *)error` in case of an error?

SystemKeeper:
> It might make sense to also add it `- (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task didCompleteWithError:(NSError *)error` in case of an error?

That's were I am not sure. When an error occurs, it should be fine to reconnect and resume a session, wouldn't it? Also, when is this called, is it also called when the app was a while in the background and the web socket fails? In this case, we should _not_  set the `disconnectTime` as this will not reflect the "real" disconnect time on the server?

PR #1712, opened 2024-07-15
Video Jump Buttons

### LINKED ISSUE #1705: Add buttons to VLC view to jump 10s back/ahead
When viewing a recording (*.webm format) we use the VLC player to display it. Currently we can only change the position in the video using the progress indicator. It would be nice if we could add buttons to jump 10s or 15s in either direction.

i2h3:
This closes #1705 by introducing buttons to jump 10 seconds backward or forward in the video playback view.

Please review as discussed earlier today, @Ivansss. 😊


Ivansss (approved):
Really nice work 👌 Thanks a lot for this PR 😊

PR #1722, opened 2024-07-23
Use MacOS 15, as 12 is now deprecated on GHA

SystemKeeper:
* Context: https://github.com/actions/runner-images/issues/10721

### DIFF
--- a/.github/workflows/uitests.yml
+++ b/.github/workflows/uitests.yml
@@ -30,13 +30,13 @@ permissions:
 
 env:
   WORKSPACE: NextcloudTalk.xcworkspace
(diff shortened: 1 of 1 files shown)


SystemKeeper:
Cool... So MacOS 14 is green, but needs to retry some tests. MacOS 13 fails. I'd suggest we stay as long on MacOS 12 as we can and then probably move to 14. Removal is planned for october this year...

erik-bershel:
It worth trying `macos-15`. Simulators seems to be 10-50% faster there.

SystemKeeper:
Thank you for the tip! Let's see :)

Ivansss (approved):

PR #1729, opened 2024-07-25
Show fallback image in failure case

SystemKeeper:
Currently when we are unable to get a preview, we fail silently and do nothing, so the activity indicator keeps spinning. In case of a failure, we should fall back to the corresponding mime type icon.

### DIFF
--- a/NextcloudTalk/BaseChatTableViewCell+File.swift
+++ b/NextcloudTalk/BaseChatTableViewCell+File.swift
@@ -119,19 +119,9 @@ extension BaseChatTableViewCell {
     // MARK: - Preview
 
     func requestPreview(for message: NCChatMessage, with account: TalkAccount) {
+        // Don't request a preview if we know that there's none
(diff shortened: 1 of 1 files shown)


NextcloudTalk/BaseChatTableViewCell+File.swift
```diff
@@ -182,9 +172,25 @@ extension BaseChatTableViewCell {
             imageView.image = image
 
             self.delegate?.cellHasDownloadedImagePreview(withHeight: ceil(previewSize.height), for: message)
+        }, failure: { _, _, _ in
+            self.showFallbackIcon(for: message)
         })
     }
 
+    func showFallbackIcon(for message: NCChatMessage) {
+        let imageName = NCUtils.previewImage(forMimeType: message.file().mimetype)
```
SystemKeeper: Missing the `-chat-preview` suffix

PR #1786, opened 2024-08-30
Continuous Voice recording

### LINKED ISSUE #618: Getting rid of the need to keep the record button pressed for voice messages
Although users highly welcome the ability to record voice messages, they completely refuse to use it on the mobile apps and still use other platforms.

The reason is that you need to have the record button pressed all the time during recording. Not only that this is annoying, in a car with the phone in a holder and connected to the handsfree this is a security issue.

All other known apps (Element, RocketChat, Signal, WhatsApp) allow to "lock" the recording button during recording the message and finally let the user decide to send or discard it.

Either the record button should be lockable (like Element, WhatsApp or Signal) or the recording should be done by starting the recording and then press "green" or "red" (like in the Web UI or like RocketChat does it - although RC is about to change this to the same way like Element, WhatsApp or Signal)

No alternative except using a different platform ;-)

rapterjet2004:
- fixes #618 

### Current Demo
https://github.com/user-attachments/assets/97d74ac4-250c-4100-aabc-c90e0694c98a

### ToDo
- [x] Added a voice recording lock button
- [x] Added a new SwiftUI View for expanded voice recording
- [x] Stop AVAudioRecorder from ending and implement the send, delete, stop/restart buttons
- [x] Figure out how to deal with layout changes on orientation and with keyboard,

### DIFF
--- a/NextcloudTalk/BaseChatViewController.swift
+++ b/NextcloudTalk/BaseChatViewController.swift
@@ -10,6 +10,7 @@ import UIKit
 import Realm
 import ContactsUI
 import QuickLook
+import SwiftUI
 
 @objcMembers public class BaseChatViewController: InputbarViewController,
                                                   UITextFieldDelegate,
@@ -80,6 +81,8 @@ import QuickLook
     private var sendButtonTagMessage = 99
     private var sendButtonTagVoice = 98
 
+    private var isVoiceRecordingLocked = false
+
     private var actionTypeTranscribeVoiceMessage = "transcribe-voice-message"
 
     private var imagePicker: UIImagePickerController?
@@ -89,6 +92,7 @@ import QuickLook
     private var voiceMessageLongPressGesture: UILongPressGestureRecognizer?
     private var recorder: AVAudioRecorder?
     private var voiceMessageRecordingView: VoiceMessageRecordingView?
+    private var expandedUIHostingController: UIHostingController<ExpandedVoiceMessageRecordingView>?
     private var longPressStartingPoint: CGPoint?
     private var cancelHintLabelInitialPositionX: CGFloat?
     private var recordCancelled: Bool = false
@@ -163,6 +167,22 @@ import QuickLook
         button.alpha = 0
         button.translatesAutoresizingMaskIntoConstraints = false
         button.setImage(UIImage(systemName: "chevron.down"), for: .normal)
+
+        self.view.addSubview(button)
+
+        return button
+    }()
+
+    private lazy var voiceRecordingLockButton: UIButton = {
+        let button = UIButton(frame: .init(x: 0, y: 0, width: 44, height: 44))
+
+        button.backgroundColor = .secondarySystemBackground
+        button.tintColor = .systemBlue
+        button.layer.cornerRadius = button.frame.size.height / 2
+        button.clipsToBounds = true
+        button.alpha = 0
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.setImage(UIImage(systemName: "lock.open"), for: .normal)
 
         self.view.addSubview(button)
 
@@ -264,7 +284,8 @@ import QuickLook
             "unreadMessageButton": self.unreadMessageButton,
             "textInputbar": self.textInputbar,
             "scrollToBottomButton": self.scrollToBottomButton,
-            "autoCompletionView": self.autoCompletionView
+            "autoCompletionView": self.autoCompletionView,
+            "voiceRecordingLockButton": self.voiceRecordingLockButton
         ]
 
         let metrics = [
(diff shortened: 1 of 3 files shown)


NextcloudTalk/BaseChatViewController.swift
```diff
@@ -170,6 +174,24 @@ import QuickLook
 
         return button
     }()
+    
+    private lazy var voiceRecordingLockButton: UIButton = {
+        let button = UIButton(frame: .init(x: 0, y: 0, width: 44, height: 44), primaryAction: UIAction { [weak self] _ in
+            // Unused atm
+        })
```
SystemKeeper:
```suggestion
        let button = UIButton(frame: .init(x: 0, y: 0, width: 44, height: 44))
```

`primaryAction` is nullable, you can just omit it, if it's not used.

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -1491,6 +1518,77 @@ import QuickLook
...
+        
+        
+        let views = [
+            "expandedVoiceMessageRecordingView": expandedVoiceMessageRecordingView
+        ]
+        
+        self.textInputbar.addConstraints(NSLayoutConstraint.constraints(withVisualFormat: "V:|[expandedVoiceMessageRecordingView]|", metrics: nil, views: views))
+        self.textInputbar.addConstraints(NSLayoutConstraint.constraints(withVisualFormat: "H:|[expandedVoiceMessageRecordingView]|", metrics: nil, views: views))
+    
+    }
```
SystemKeeper:
```suggestion
    func showExpandedVoiceMessageRecordingView(offset: Int) {
        let expandedView = ExpandedVoiceMessageRecordingView(
            deleteFunc: handleDelete, sendFunc: handleSend, recordFunc: handleRecord(isRecording:), timeElapsed: offset
        )

        let hostingController = UIHostingController(rootView: expandedView)
        guard let expandedVoiceMessageRecordingView = hostingController.view else { return }

        self.expandedUIHostingController = hostingController
        self.textInputbar.addSubview(expandedVoiceMessageRecordingView)

        expandedVoiceMessageRecordingView.translatesAutoresizingMaskIntoConstraints = false

        let views = [
            "expandedVoiceMessageRecordingView": expandedVoiceMessageRecordingView
        ]

        self.textInputbar.addConstraints(NSLayoutConstraint.constraints(withVisualFormat: "V:|[expandedVoiceMessageRecordingView]|", metrics: nil, views: views))
        self.textInputbar.addConstraints(NSLayoutConstraint.constraints(withVisualFormat: "H:|[expandedVoiceMessageRecordingView]|", metrics: nil, views: views))
    }
```

This way you don't have to force-unwrap anything.

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -1491,6 +1518,77 @@ import QuickLook
...
+    
+    func handleDelete() {
+        print("Delete Handled")
```
SystemKeeper: If you want to keep these prints, could you make them a bit more explicit, so it is clear that it's from an audio message?

SystemKeeper (commented):
Rotating the screen breaks the UI:
Simulator Screenshot - iPhone 15 Pro - 2024-09-08 at 20 36 42

Also I noticed that the buttons are not clickable in the bottom area?! 🤔 

Please also take care of the swiftlint warnings :)

SystemKeeper:
Really nice work otherwise, thanks a lot 👍

Ivansss (changes requested):
Thanks a lot @rapterjet2004 for implementing this nice feature! :)

The PR overall looks good, but there is an issue when the user tries to lock the voice recording when the keyboard is being shown.
The new view to delete/stop/send the recording is hidden under the keyboard and it only appears when hidding the keyboard.


https://github.com/user-attachments/assets/fc25760c-851c-4291-9aa8-0cf940f29bb7

SystemKeeper:
Something is still weird when trying to click the stop button, it does not work for me:

https://github.com/user-attachments/assets/2cd86cd2-ddb2-49ba-9afb-d2d5159b7ab3

Deleting and sending does work?

Seems the padding is not correct as some icons are covered now:

Bildschirmfoto 2024-10-09 um 22 45 43


I didn't test too much, but maybe something like this is what you want to achieve?

```diff
diff --git a/NextcloudTalk/BaseChatViewController.swift b/NextcloudTalk/BaseChatViewController.swift
index b960f5ad..2b94fe45 100644
--- a/NextcloudTalk/BaseChatViewController.swift
+++ b/NextcloudTalk/BaseChatViewController.swift
@@ -1528,7 +1528,7 @@ import SwiftUI
         guard let expandedVoiceMessageRecordingView = hostingController.view else { return }
 
         self.expandedUIHostingController = hostingController
-        self.textInputbar.addSubview(expandedVoiceMessageRecordingView)
+        self.view.addSubview(expandedVoiceMessageRecordingView)
 
         expandedVoiceMessageRecordingView.translatesAutoresizingMaskIntoConstraints = false
 
@@ -1536,7 +1536,7 @@ import SwiftUI
             "expandedVoiceMessageRecordingView": expandedVoiceMessageRecordingView
         ]
 
-        self.view.addConstraints(NSLayoutConstraint.constraints(withVisualFormat: "V:|-(>=0)-[expandedVoiceMessageRecordingView]|", metrics: nil, views: views))
+        expandedVoiceMessageRecordingView.bottomAnchor.constraint(equalTo: self.textInputbar.bottomAnchor).isActive = true
         self.view.addConstraints(NSLayoutConstraint.constraints(withVisualFormat: "H:|[expandedVoiceMessageRecordingView]|", metrics: nil, views: views))
     }
```

SystemKeeper:
> but maybe something like this is what you want to achieve?

You can do a similar thing for left and right anchor as well, if you don't like the visual layout format.

SystemKeeper (approved):
I still see a few corner-cases here and there, but I think this is fine to do at some point in another PR. From my POV fine to merge. Thanks for the PR 👍

Ivansss (approved):
Thanks a lot for your work here @rapterjet2004 :)
As @SystemKeeper said, there are small UI issues that could be tackled in a follow-up PR ;)

PR #1799, opened 2024-09-11
Return localVideoView to original position after being pushed by right side bar

Ivansss:
**Before this PR:**
The localVideoView is pushed to the middle of the screen when the right side bar is being shown.
When the side bar is closed, the localVideoView stays at the same position where it was pushed.

**After this PR:**
The localVideoView will return to right side when the right side bar is getting closed.


SystemKeeper (approved):
Tested and works. The only downside I see is that the location will change as soon as you're in the side-bars widths, but I don't think this will really be an issue

PR #1801, opened 2024-09-13
Improve zooming in NCZoomableView

Ivansss:
Add pinch and pan gestures recognizers to zoomableView instead of contentView so images/videos can be zoomed and dragged even if the gesture started outside the image/video frame.
This improves the zooming experience.


SystemKeeper (approved):
Tested and works

PR #1814, opened 2024-09-22
Ensure non-nil sessionId

SystemKeeper:
* Follow-up to https://github.com/nextcloud/talk-ios/pull/1789

Fixed in this PR:
* SessionId is an optional, we only rely on the HTTP status code currently
* For creating a room object, we only rely on the listable capability currently
* In ChatViewController we just ensure that there's a NCRoom object, not, that it is the correct one

(The last 2 fixes are fixing basically the same thing, but I think we should be explicit here)

### DIFF
--- a/NextcloudTalk/ChatViewController.swift
+++ b/NextcloudTalk/ChatViewController.swift
@@ -202,10 +202,6 @@ import SwiftyAttributes
         }
 
         if !self.offlineMode {
-            if self.room.token == nil {
-                fatalTokenError()
-            }
-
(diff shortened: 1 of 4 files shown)


SystemKeeper:
Most of the observed crashes that were observed with a null-token seem to have gone, which makes we wonder if they were really a bug in our (talk/talk-ios) codebase...
Therefore I think we can remove the `fatalError` experiment again (didn't yield any results anyway). Still I would like to add the safe-guards, just in case.

Ivansss (approved):

PR #1386, opened 2023-10-15
Migrate NCChatViewController to Swift

SystemKeeper:
This PR migrates our `NCChatViewController` to Swift and tries to decouple its responsibilities. 

With this PR we have 3 ViewControllers:

#### `InputbarViewController` 
* Subclass of `SLKTextViewController`
* Responsible for the appearance of the ViewControllers and the TextInputbar.
* Contains everything needed for a TextInputbar and AutoCompletion
* Allows initialization with a UIView instaed of a UITableView (Note: The constraints are not finished yet, but generally it's working)

Snippet to test:
```objc
    NCRoom *lastRoom = [NCRoom allObjects].lastObject;
    NCRoom *lastRoomUnmanaged = [[NCRoom alloc] initWithValue:lastRoom];
    UIView *testView = [[UIView alloc] init];
    [testView setBackgroundColor:[UIColor systemRedColor]];
    InputbarViewController *inputbarVC = [[InputbarViewController alloc] initFor:lastRoomUnmanaged withView:testView];
    NCNavigationController *navigationController = [[NCNavigationController alloc] initWithRootViewController:inputbarVC];
    [self presentViewController:navigationController animated:YES completion:nil];
```

#### `BaseChatViewController`
* Subclass of `InputbarViewController`
* Responsible for everything chat related, that does not need a ChatController, e.g. cell-rendering, cell height calculation, etc.
* Allows initialization with an Array of `NCChatMessage` for displaying arbitrary chat messages (e.g. when selecting a message in a search result and then obtaining the context of that message)

Snippet to test:
```objc
    NCRoom *lastRoom = [NCRoom allObjects].lastObject;
    NCRoom *lastRoomUnmanaged = [[NCRoom alloc] initWithValue:lastRoom];

    NSPredicate *query = [NSPredicate predicateWithFormat:@"accountId = %@ AND token = %@", lastRoomUnmanaged.accountId, lastRoomUnmanaged.token];

    RLMResults *managedMessages = [NCChatMessage objectsWithPredicate:query];
    RLMResults *managedSortedMessages = [managedMessages sortedResultsUsingKeyPath:@"messageId" ascending:YES];
    NSMutableArray *sortedMessages = [NSMutableArray new];

    for (NSInteger i = 0; i < 90; i++) {
        NCChatMessage *sortedMessage = [[NCChatMessage alloc] initWithValue:managedSortedMessages[i]];
        [sortedMessages addObject:sortedMessage];
    }

[...]

PR #1817, opened 2024-09-26
New audio chat messages

Ivansss:
- [x] Swift
- [x] Fixes regression introduced in https://github.com/nextcloud/talk-ios/pull/1386 
        Not being able to pause and continue voice messages playback.

Before:

IMG_08B14F9F8F5E-1

After:

IMG_0066


SystemKeeper (approved):
Really nice work 👍

PR #1824, opened 2024-10-01
User status message setting fixes

### LINKED ISSUE #1823: Unable to set status message without emoji
Try to just change your status message, without selecting an emoji. Talk-ios will always use the default emoji, while on web it is possible.

Ivansss:
- [x] Allow to set user status with only an icon or only a message
- [x] Enable "Clear status message" only when the user has one status already set

Fix #1823


Black-Fox-2022 (approved):

SystemKeeper (approved):
Nice work 👍

PR #1838, opened 2024-10-16
Register poll message cell identifiers in BaseChatViewController table view

Ivansss:
Since those identifiers were not registered `chatMessageCellIdentifier` were being used for poll messages


SystemKeeper (approved):
Ah it happened after the last change probably 😅

PR #1839, opened 2024-10-16
Convert extensions to build folders

SystemKeeper:
Convert our extensions to build folders instead of groups. This only adds the folders instead of individual files and should (when we finish it) result in less changes to the project file.


SystemKeeper:
CocoaPods not happy 😓

SystemKeeper:
Error comes from outdated CocoaPods version on the runners, see `github.com/actions/runner-images/issues/10874`

Ivansss (approved):

PR #1844, opened 2024-10-18
Support videos in media view

### LINKED ISSUE #1702: Support videos in media view
Currently when opening a picture and "swiping" through the list of recieved images, videos are skipped. It would be nice if we would show them with a play button there as well. This view:

Question:
When opening a video, should that be opened in that media view then as well?

DanielStandfest:
added support for videos (and playback) when swiping in media view.

Closes #1702 

@SystemKeeper

### DIFF
--- a/NextcloudTalk/BaseChatViewController.swift
+++ b/NextcloudTalk/BaseChatViewController.swift
@@ -3378,6 +3378,20 @@ import SwiftUI
             return
         }
 
+        let filePath = fileParameter.path ?? ""
+        let fileExtension = URL(fileURLWithPath: filePath).pathExtension.lowercased()
+
+        if NCUtils.isVideo(fileType: fileParameter.mimetype) {
+            // Skip unsupported formats here ("webm" and "mkv") and use VLC later
+            if !fileExtension.isEmpty, !VLCKitVideoViewController.supportedFileExtensions.contains(fileExtension) {
+                let mediaViewController = NCMediaViewerViewController(initialMessage: message)
+                let navController = CustomPresentableNavigationController(rootViewController: mediaViewController)
+
+                self.present(navController, interactiveDismissalType: .standard)
+                return
+            }
+        }
+
         if fileParameter.fileStatus != nil && fileParameter.fileStatus?.isDownloading ?? false {
             print("File already downloading -> skipping new download")
             return
--- a/NextcloudTalk/NCMediaViewerPageViewController.swift
+++ b/NextcloudTalk/NCMediaViewerPageViewController.swift
@@ -3,13 +3,15 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 //
 
+import AVKit
+import AVFoundation
 import Foundation
 import UIKit
 import SwiftyGif
 
 @objc protocol NCMediaViewerPageViewControllerDelegate {
     @objc func mediaViewerPageZoomDidChange(_ controller: NCMediaViewerPageViewController, _ scale: Double)
-    @objc func mediaViewerPageImageDidLoad(_ controller: NCMediaViewerPageViewController)
+    @objc func mediaViewerPageMediaDidLoad(_ controller: NCMediaViewerPageViewController)
 }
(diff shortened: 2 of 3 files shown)


SystemKeeper:
Very cool @DanielStandfest, thanks a lot for working on this 👍
We are going to take a look at this in the upcoming week. 

Btw, we can add you to the Nextcloud org if you like? Then you can directly push here. We also have a developer chat where we can create a guest account for you, if you like. Just let me know :)

DanielStandfest:
Thank you @SystemKeeper.
Sure, would be nice to be added :)

Appreciate it.

SystemKeeper:
The warning about the branch being out of date is usually not an issue, it's just a warning. If merging would not be possible
you would see a message like:
image

Since we pull in latest translation changes at least once a day, it's very likely that your branch becomes outdated quite quickly.

If you wanna update the branch to be inline with latest master, you can use the rebase option to prevent unnecessary merge commits btw:
image

DanielStandfest:
Any news on this one? 

@SystemKeeper

Ivansss:
Hello @DanielStandfest ,

Thanks a lot for your PR :) 

Sorry for our late reply, @SystemKeeper and I did take a look to your PR last week but forgot to write a reply here.

We noticed a few things that could be addressed before merging your PR:

1. It would be nice if the NCMediaViewerViewController is opened directly when tapping on a video file too:
https://github.com/nextcloud/talk-ios/blob/ba727cd6e011251d9d8bbeaede1182a1cb6a855d/NextcloudTalk/BaseChatViewController.swift#L3312-L3319

2. We could skip for now "webm" and "mkv" videos and do not try to present them in the NCMediaViewerViewController and just present them using VLC:
https://github.com/nextcloud/talk-ios/blob/ba727cd6e011251d9d8bbeaede1182a1cb6a855d/NextcloudTalk/BaseChatViewController.swift#L3536

3. (Optional for this PR, could be done in a follow-up PR by us or you)
Instead of directly downloading videos when presenting them in the NCMediaViewerViewController, we could show a play button and the preview of the video (if available) and only when tapping on the play button we start downloading the video

DanielStandfest:
Hi @Ivansss,

thank you for your valuable feedback!

Points 1 and 2 have now been implemented. Additionally, I noticed the issue with unsupported video formats when swiping through webm and mkv videos. I’ve adjusted the getPreviousFileMessage and getNextFileMessage methods to address this.

Regarding point 3, I’d suggest handling it as a separate feature in a future update, and I recommend creating a dedicated issue to track it.

Thanks again for your insights!

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -3318,6 +3318,20 @@ import SwiftUI
             return
         }
 
+        let fileLocalPath = fileParameter.fileStatus?.fileLocalPath ?? ""
+        let fileExtension = URL(fileURLWithPath: fileLocalPath).pathExtension.lowercased()
```
Ivansss:
```suggestion
        let filePath = fileParameter.path ?? ""
        let fileExtension = URL(fileURLWithPath: filePath).pathExtension.lowercased()
```

`fileStatus` is only available when we start downloading a file, that's why at the moment when you tap on a video the first time, it will start downloading the video file and it will open it in the QLPreviewController. Then the next times it will open it directly in the NCMediaViewer, since it can get the local path from the fileStatus.

If we get the `path` directly from the fileParameter, we are able to check its extension and the video file will be opened in the NCMediaViewer directly.

Ivansss (changes requested):

NextcloudTalk/BaseChatViewController.swift
DanielStandfest: implemented

SystemKeeper (approved):
Very nice @DanielStandfest ! Thanks for your contribution 🥳

Ivansss (approved):
Thanks a lot for your PR @DanielStandfest :)

PR #1846, opened 2024-10-21
Add new TextFieldTableViewCell and TextViewTableViewCell that use a dynamic font

Ivansss:


NextcloudTalk/RoomAvatarInfoTableViewController.swift
```diff
@@ -126,8 +126,8 @@ enum RoomAvatarInfoSection: Int {
     }
 
     override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        let textInputCell = tableView.dequeueReusableCell(withIdentifier: kTextInputCellIdentifier) as? TextInputTableViewCell ??
-        TextInputTableViewCell(style: .default, reuseIdentifier: kTextInputCellIdentifier)
+        let textInputCell = tableView.dequeueReusableCell(withIdentifier: textFieldCellIdentifier) as? TextFieldTableViewCell ??
+        TextFieldTableViewCell(style: .default, reuseIdentifier: textFieldCellIdentifier)
```
SystemKeeper: I think if you pass the indexPath to the dequeue here, you are guaranteed to get a cell and can remove the coalesce part

NextcloudTalk/RoomCreationTableViewController.swift
```diff
@@ -426,8 +426,8 @@ enum RoomVisibilityOption: Int {
         let roomCreationSection = sections[indexPath.section]
 
         if roomCreationSection == RoomCreationSection.kRoomNameSection.rawValue {
-            let textInputCell = tableView.dequeueReusableCell(withIdentifier: kTextInputCellIdentifier) as? TextInputTableViewCell ??
-            TextInputTableViewCell(style: .default, reuseIdentifier: kTextInputCellIdentifier)
+            let textInputCell = tableView.dequeueReusableCell(withIdentifier: textFieldCellIdentifier) as? TextFieldTableViewCell ??
+            TextFieldTableViewCell(style: .default, reuseIdentifier: textFieldCellIdentifier)
```
SystemKeeper: Same as above

NextcloudTalk/UserProfileTableViewController.swift
```diff
@@ -261,9 +261,9 @@ extension UserProfileTableViewController {
 
     // MARK: Setup cells
 
-    func textInputCellWith(text: String?, tag: Int?, interactionEnabled: Bool?, keyBoardType: UIKeyboardType?, autocapitalizationType: UITextAutocapitalizationType?, placeHolder: String?) -> TextInputTableViewCell {
-        let textInputCell = tableView.dequeueReusableCell(withIdentifier: kTextInputCellIdentifier) as? TextInputTableViewCell ??
-        TextInputTableViewCell(style: .default, reuseIdentifier: kTextInputCellIdentifier)
+    func textInputCellWith(text: String?, tag: Int?, interactionEnabled: Bool?, keyBoardType: UIKeyboardType?, autocapitalizationType: UITextAutocapitalizationType?, placeHolder: String?) -> TextFieldTableViewCell {
+        let textInputCell = tableView.dequeueReusableCell(withIdentifier: textFieldCellIdentifier) as? TextFieldTableViewCell ??
+        TextFieldTableViewCell(style: .default, reuseIdentifier: textFieldCellIdentifier)
```
SystemKeeper: Same :)

NextcloudTalk/PollCreationViewController.swift
```diff
@@ -192,41 +192,41 @@ import UIKit
...
             textInputCell.textField.tag = kQuestionTextFieldTag
             textInputCell.textField.text = question
             return textInputCell
         } else if indexPath.section == PollCreationSection.kPollCreationSectionOptions.rawValue {
             if indexPath.row == options.count {
+                let actionCell = tableView.dequeueOrCreateCell(withIdentifier: "PollSettingCellIdentifier")
                 actionCell.textLabel?.text = NSLocalizedString("Add answer", comment: "")
                 return actionCell
             } else if indexPath.row < options.count {
+                let textInputCell: TextFieldTableViewCell = tableView.dequeueOrCreateCell(withIdentifier: textFieldCellIdentifier)
```
SystemKeeper: The delegate is missing here, resulting in non-trimmed answeres. Also autocapitalizationType is missing, but since it's the default value anyway (https://developer.apple.com/documentation/uikit/uitextinputtraits/1624447-autocapitalizationtype) it doesn't make a difference.

Edit: Ah no, since you reset the type in `prepareForReuse`, we need to set it here again, as otherwise we loose the autocap. functionality. Which makes me wonder, should reuse and init at least do the same?

NextcloudTalk/PollCreationViewController.swift
```diff
@@ -192,41 +192,41 @@ import UIKit
...
                 actionCell.textLabel?.text = NSLocalizedString("Multiple answers", comment: "")
                 actionCell.accessoryView = multipleSwitch
                 return actionCell
             }
-            return actionCell
+            return UITableViewCell()
         }
-
-        return actionCell
+        return UITableViewCell()
```
SystemKeeper: Just keep the last `return UITableViewCell()` here?

NextcloudTalk/TextFieldTableViewCell.swift
```diff
@@ -0,0 +1,48 @@
+//
+// SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+// SPDX-License-Identifier: GPL-3.0-or-later
+//
+
+import UIKit
+
+let textFieldCellIdentifier = "textFieldCellIdentifier"
```
SystemKeeper: This should be better inside the class, as it's done in `TextViewTableViewCell`? 🤔

SystemKeeper (changes requested):

SystemKeeper (approved):

PR #1851, opened 2024-10-23
fixed SettingsViewController not updating after clearing caches

### LINKED ISSUE #1791: SettingsViewController not updated after clearing caches
## Steps to reproduce
1. go to settings 
2. Clear a cache

### Expected behaviour
The displayed amount should be updated

DanielStandfest:
Fixes #1791

@SystemKeeper

### DIFF
--- a/NextcloudTalk/SettingsTableViewController.swift
+++ b/NextcloudTalk/SettingsTableViewController.swift
@@ -120,18 +120,8 @@ class SettingsTableViewController: UITableViewController, UITextFieldDelegate, U
         NotificationCenter.default.addObserver(self, selector: #selector(contactsAccessHasBeenUpdated(notification:)), name: NSNotification.Name.NCContactsManagerContactsAccessUpdated, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(userProfileImageUpdated), name: NSNotification.Name.NCUserProfileImageUpdated, object: nil)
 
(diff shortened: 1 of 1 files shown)


SystemKeeper:
Hey, thanks again for this PR!
I actually worked on that as well in this PR https://github.com/nextcloud/talk-ios/pull/1849. But I missed the completion block for the disk cache, which is nice. So I'll remove that commit from the PR and we can go ahead with yours :)

The only thing I am wondering: Did you see any threading issues? Everything should be done on the main thread already, the completion block for example is explicitly dispatched to main already:
https://github.com/SDWebImage/SDWebImage/blob/780aa6db621b7d195c31bab4d483879c1290d278/SDWebImage/Core/SDImageCache.m#L822-L831

That's the only thing I am wondering right now (and is a bit nit-picky 😉).

DanielStandfest:
Hi, thank you for your review, Marcel.
Funny that we worked on the same issue at (nearly) the same time.

No problem with being nit-picky, I am eager to learn!
You are right and it seems that there are no threading issues after removing the dispatchqueue-main thread block.

I removed these parts now :)

SystemKeeper (approved):
Cool, thanks for the PR and the changes :-)

Ivansss (approved):
Tested and works 👍 
Thanks a lot for your PR @DanielStandfest :)

PR #1872, opened 2024-11-10
Increase layer speed on UI test

SystemKeeper:
Leftover from https://github.com/nextcloud/talk-ios/pull/1550


SystemKeeper:
Nvm. I don't think it's worth it this time.

Issue #1261, opened 2023-06-01
Better voice message handling


e-caste:
**Is your feature request related to a problem? Please describe.**
Voice messages work but could be handled better in my opinion.

**Describe the solution you'd like**
A few things I've noticed, in order of importance:
- there is no sign that indicates whether the voice message that has just been recorded is being sent to the server or discarded, the user just has to cross their fingers and wait until it appears (or not) in the chat
- if a voice message can't be sent to the server (e.g. you've switched 4/5G<->WiFi), it would be great to be able to re-send it without having to re-record it from scratch
- the scrubbing doesn't work well when voice messages are playing, they first have to be stopped (and even then, it's hard to grab the round "handle" to move it, I suspect the active area should be made larger)
- almost all messaging apps currently implement some form of different playback speeds, good options would be 0.5x, 1x, 1.5x, 2x in my opinion. This is missing at the moment
- related to the above: Telegram and Whatsapp have recently introduced an additional bar at the top that shows if a voice message is being played, which is permanent throughout the chats and allows to toggle between playback speeds, would be nice to have to be able to read other chats while listening to a voice message 

**Describe alternatives you've considered**
Not applicable.

**Additional context**
Talk iOS 16.0.2.

PR #1874, opened 2024-11-10
added temporary voice messages

DanielStandfest:
added support for temporary voice messages and handling them just like text messages.

related to #1261

Would be happy to get feedback! :)

Cheers,
Daniel


DanielStandfest:
I am not sure why the localizable change is not passing the test? :/

SystemKeeper:
> I am not sure why the localizable change is not passing the test? :/

Interesting, seems to be a sorting issue. I’ll take a look.

SystemKeeper:
> I’ll take a look.

So I tried multiple things yesterday and it only happens if you rebase on master (that is what CI does and that's why you can't seem to trigger that change locally in your branch). I have the same issue in https://github.com/nextcloud/talk-ios/pull/1880. I couldn't fully figure out what's triggering this, even rebasing on old commits did not change that. So for now I'd say we ignore it and take a look later when the PR is merged. 

We are going to have a closer look to this and your other PR this week :)

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -475,14 +475,41 @@ import QuickLook
         temporaryMessage.actorType = "users"
         temporaryMessage.timestamp = Int(Date().timeIntervalSince1970)
         temporaryMessage.token = room.token
-        temporaryMessage.message = self.replaceMentionsDisplayNamesWithMentionsKeysInMessage(message: message, parameters: messageParameters)
 
         let referenceId = "temp-\(Date().timeIntervalSince1970 * 1000)"
         temporaryMessage.referenceId = NCUtils.sha1(fromString: referenceId)
         temporaryMessage.internalId = referenceId
         temporaryMessage.isTemporary = true
+        temporaryMessage.isOfflineMessage = true
```
SystemKeeper: `isOfflineMessage` indicates that this message was sent without an internet connection and should be resend, when connection is available again. So this shouldn't be added here.
There's quite some logic involved (including trying to resend in the background). Check out e.g. https://github.com/nextcloud/talk-ios/blob/dd3e20b77101a741aecff36f624c0fe7ab79a7c7/NextcloudTalk/NCChatController.m#L685-L685

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -475,14 +475,41 @@ import QuickLook
...
         temporaryMessage.parentId = parentMessage?.internalId
-        temporaryMessage.messageParametersJSONString = messageParameters
```
SystemKeeper, line 525: This is lost for normal messages, resulting in broken mentions: 
image

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -1590,52 +1634,82 @@ import QuickLook
         audioFileName += ".mp3"
 
         let activeAccount = NCDatabaseManager.sharedInstance().activeAccount()
+        NCAPIController.sharedInstance().setupNCCommunication(for: activeAccount)
+        let chatFileController = NCChatFileController()
+        chatFileController.initDownloadDirectory(for: activeAccount)
+
+        let destinationFilePath = "\(chatFileController.tempDirectoryPath)/\(audioFileName)"
```
SystemKeeper: Usually it's better to leave appending filesnames/directories to methods like `appendingPathComponent`

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -1590,52 +1634,82 @@ import QuickLook
...
+            silently: false,
+            isVoiceMessage: true
+        )
+
+        chatFileController.setFileStatusFrom(temporaryMessage.file())
+
+        _ = chatFileController.moveFileToTemporaryDirectory(
+            fromSourcePath: self.recorder!.url.path,
+            destinationPath: destinationFilePath
+        )
```
SystemKeeper: If you don't care about the methods return value, why having a return value in the first place?
I guess if we can't move it, we should not show a temporary message?

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -475,14 +475,41 @@ import QuickLook
...
+                "type": "file",
+                "name": message,
+                "path": messageParameters,
+                "fileId": parameterId,
+                "fileName": message,
+                "filePath": messageParameters,
+                "fileLocalPath": messageParameters
+            ]
+
+            messageParametersDict[parameterId] = fileParameterDict
```
SystemKeeper: Why using a UUID here? Should it be the file dictionary? I don't see it used anywhere? (Maybe I missed it, not yet through everything).

SystemKeeper (commented):
I haven't looked through everything, here are some first comments :)

NextcloudTalk/BaseChatViewController.swift
DanielStandfest: Fixed

NextcloudTalk/BaseChatViewController.swift
DanielStandfest: Good question, I guess I had a problem with the file parameter not being set when putting the fileParameters in no dictionary.
What's your suggestion?

NextcloudTalk/BaseChatViewController.swift
DanielStandfest: True, fixed.

NextcloudTalk/BaseChatViewController.swift
DanielStandfest: Good point, should be fixed now.

NextcloudTalk/BaseChatViewController.swift
DanielStandfest: Thank you for the information, I recognized that the context menu fits better when using the isOfflineMessage = true. That's why I set it to true. 
For example the "open in nextcloud" (shouldn't really be an option), "copy" and "create reminder" option dont work out of the box.

I removed it now, should we fix some of the options like "delete" then or hide some of the invalid options?

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -1590,52 +1634,83 @@ import QuickLook
         audioFileName += ".mp3"
 
         let activeAccount = NCDatabaseManager.sharedInstance().activeAccount()
+        NCAPIController.sharedInstance().setupNCCommunication(for: activeAccount)
```
SystemKeeper: Was there a reason to setup the communication here? Because in `uniqueNameForFileUploadWithName` there's already a setup call and I don't think it's needed before that call.

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -1590,52 +1634,83 @@ import QuickLook
...
+
+        let temporaryMessage = self.createTemporaryMessage(
+            message: audioFileName,
+            replyTo: nil,
+            messageParameters: "\(destinationFilePath)",
+            silently: false,
+            isVoiceMessage: true
+        )
+
+        chatFileController.setFileStatusFrom(temporaryMessage.file())
```
SystemKeeper: Since we are not using the `chatFileController` in this method again (besides moving the file), we don't need the filestatus to be set.

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -1590,52 +1634,83 @@ import QuickLook
...
+
+        let movedFileToTemporaryDirectory = chatFileController.moveFileToTemporaryDirectory(
+            fromSourcePath: self.recorder!.url.path,
```
SystemKeeper: Please use something like `guard let recorder = self.recorder else { return }` at the beginning of the file, so we don't have to force-unwrap here.

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -1590,52 +1634,83 @@ import QuickLook
...
+        if movedFileToTemporaryDirectory, NCDatabaseManager.sharedInstance().roomHasTalkCapability(kCapabilityChatReferenceId, for: room) {
+            self.appendTemporaryMessage(temporaryMessage: temporaryMessage)
+        }
+
         NCAPIController.sharedInstance().uniqueNameForFileUpload(withName: audioFileName, originalName: true, for: activeAccount, withCompletionBlock: { fileServerURL, fileServerPath, _, _ in
             if let fileServerURL, let fileServerPath, let recorder = self.recorder {
                 let talkMetaData: [String: String] = ["messageType": "voice-message"]
-                self.uploadFileAtPath(localPath: recorder.url.path, withFileServerURL: fileServerURL, andFileServerPath: fileServerPath, withMetaData: talkMetaData)
+
+                self.uploadFileAtPath(localPath: recorder.url.path, withFileServerURL: fileServerURL, andFileServerPath: fileServerPath, withMetaData: talkMetaData, temporaryMessage: temporaryMessage)
```
SystemKeeper:
```suggestion
                self.uploadFileAtPath(localPath: destinationFilePath, withFileServerURL: fileServerURL, andFileServerPath: fileServerPath, withMetaData: talkMetaData, temporaryMessage: temporaryMessage)
```

Otherwise the file is not found on upload (because we moved it before) and we get a 0 byte file.

NextcloudTalk/ChatFileUploader.swift
```diff
@@ -0,0 +1,63 @@
...
+                                                                   talkMetaData: talkMetaData,
+                                                                   referenceId: temporaryMessage?.referenceId) { shareError in
+                    if let shareError = shareError {
+                        NSLog("Failed to share voice message: \(shareError.localizedDescription)")
+                        completion(403, "Failed to share voice message")
+                    } else {
+                        completion(200, nil)
+                    }
+                }
+            case 404:
```
SystemKeeper: In the original code, also case 409 was checked here.

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -1590,52 +1634,83 @@ import QuickLook
...
+                switch statusCode {
+                case 200:
+                    NSLog("Successfully uploaded and shared voice message")
+                case 401:
+                    NSLog("No active account found")
+                    self.showAlert(title: NSLocalizedString("Upload failed", comment: ""), message: NSLocalizedString("No active account found", comment: ""))
+                case 403:
+                    NSLog("Failed to share voice message")
+                    self.showAlert(title: NSLocalizedString("Upload failed", comment: ""), message: NSLocalizedString("Failed to share recording", comment: ""))
+                case 404:
```
SystemKeeper: Also needs to be checked for 409.

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -1590,52 +1634,83 @@ import QuickLook
...
+                }
             }
-        })
+        }
+    }
+
+    func showAlert(title: String, message: String) {
+        let alert = UIAlertController(title: title, message: message, preferredStyle: .alert)
+        alert.addAction(UIAlertAction(title: NSLocalizedString("OK", comment: ""), style: .default))
+        self.present(alert, animated: true)
```
SystemKeeper: We have https://github.com/nextcloud/talk-ios/blob/1602e8fb25a91ad0fb6bfa3ad3a0c59cba0d66ec/NextcloudTalk/NCUserInterfaceController.m#L323-L331 maybe you could reuse that?

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -3236,12 +3311,22 @@ import QuickLook
...
+            return
+        }
+
         if fileParameter.fileStatus != nil && fileParameter.fileStatus?.isDownloading ?? false {
             print("File already downloading -> skipping new download")
             return
         }
 
+        if fileParameter.fileStatus?.fileLocalPath != nil && FileManager.default.fileExists(atPath: fileParameter.fileStatus?.fileLocalPath ?? "") {
+            self.setupVoiceMessagePlayer(with: fileParameter.fileStatus!)
```
SystemKeeper:
```suggestion
        if let fileStatus = fileParameter.fileStatus, fileStatus.fileLocalPath != nil && FileManager.default.fileExists(atPath: fileStatus.fileLocalPath ?? "") {
            self.setupVoiceMessagePlayer(with: fileStatus)
```

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -1590,52 +1634,83 @@ import QuickLook
...
+            fromSourcePath: self.recorder!.url.path,
+            destinationPath: destinationFilePath
+        )
+
+        if movedFileToTemporaryDirectory, NCDatabaseManager.sharedInstance().roomHasTalkCapability(kCapabilityChatReferenceId, for: room) {
+            self.appendTemporaryMessage(temporaryMessage: temporaryMessage)
+        }
+
         NCAPIController.sharedInstance().uniqueNameForFileUpload(withName: audioFileName, originalName: true, for: activeAccount, withCompletionBlock: { fileServerURL, fileServerPath, _, _ in
             if let fileServerURL, let fileServerPath, let recorder = self.recorder {
```
SystemKeeper:
```suggestion
            if let fileServerURL, let fileServerPath {
```

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -1590,52 +1634,83 @@ import QuickLook
...
+        )
+
+        chatFileController.setFileStatusFrom(temporaryMessage.file())
+
+        let movedFileToTemporaryDirectory = chatFileController.moveFileToTemporaryDirectory(
+            fromSourcePath: self.recorder!.url.path,
+            destinationPath: destinationFilePath
+        )
+
+        if movedFileToTemporaryDirectory, NCDatabaseManager.sharedInstance().roomHasTalkCapability(kCapabilityChatReferenceId, for: room) {
```
SystemKeeper: We can also use the `movedFileToTemporaryDirectory` in a guard, because we don't want to upload a file, when the moving did not work?

NextcloudTalk/BaseChatViewController.swift
```diff
@@ -1590,52 +1634,83 @@ import QuickLook
...
+                case 404:
+                    NSLog("Failed to check or create attachment folder")
+                    self.showAlert(title: NSLocalizedString("Upload failed", comment: ""), message: NSLocalizedString("Failed to check or create attachment folder", comment: ""))
+                case 507:
+                    NSLog("User storage quota exceeded")
+                    let alert = UIAlertController(title: NSLocalizedString("Upload failed", comment: ""),
+                                                  message: NSLocalizedString("User storage quota exceeded", comment: ""),
+                                                  preferredStyle: .alert)
+                    alert.addAction(UIAlertAction(title: NSLocalizedString("OK", comment: ""), style: .default))
+                    self.present(alert, animated: true)
```
SystemKeeper: Also use `showAlert` (or the other method, see below) here? :)

NextcloudTalk/NCChatFileController.h
```diff
@@ -28,7 +28,11 @@ extern NSString * const NCChatFileControllerDidChangeDownloadProgressNotificatio
 @property (nonatomic, weak) id<NCChatFileControllerDelegate> delegate;
 @property (nonatomic, strong) NSString *messageType;
 @property (nonatomic, strong) NSString *actionType;
+@property (nonatomic, strong, readonly) NSString *tempDirectoryPath;
 
+- (void)initDownloadDirectoryForAccount:(TalkAccount *)account;
+- (void)setFileStatusFromParameter:(NCMessageFileParameter *)fileParameter;
```
SystemKeeper: If I am not mistaken, this is not needed anymore, since we don't call it?!

NextcloudTalk/NCChatFileController.m
```diff
@@ -172,6 +171,39 @@ - (void)downloadFileWithFileId:(NSString *)fileId
     }];
 }
 
+- (void)setFileStatusFromParameter:(NCMessageFileParameter *)fileParameter {
+    _fileStatus = [[NCChatFileStatus alloc] initWithFileId:fileParameter.parameterId fileName:fileParameter.name filePath:fileParameter.path fileLocalPath:fileParameter.fileStatus.fileLocalPath];
+    fileParameter.fileStatus = _fileStatus;
```
SystemKeeper: So also not needed?

NextcloudTalk/NCChatFileController.m
```diff
@@ -172,6 +171,39 @@ - (void)downloadFileWithFileId:(NSString *)fileId
...
+
+    if ([[NSFileManager defaultManager] fileExistsAtPath:destinationPath]) {
+        NSLog(@"File is already in temporary directory: %@", destinationPath);
+        return false;
+    }
+
+    NSError *error = nil;
+    if ([[NSFileManager defaultManager] moveItemAtPath:sourcePath toPath:destinationPath error:&error]) {
+        NSLog(@"File successfully moved to: %@", destinationPath);
+        _fileStatus.fileLocalPath = destinationPath;
```
SystemKeeper: Since the fileStatus is not used, we can remove that line as well

NextcloudTalk/NCChatFileController.m
```diff
@@ -172,6 +171,39 @@ - (void)downloadFileWithFileId:(NSString *)fileId
...
 
+- (void)setFileStatusFromParameter:(NCMessageFileParameter *)fileParameter {
+    _fileStatus = [[NCChatFileStatus alloc] initWithFileId:fileParameter.parameterId fileName:fileParameter.name filePath:fileParameter.path fileLocalPath:fileParameter.fileStatus.fileLocalPath];
+    fileParameter.fileStatus = _fileStatus;
+}
+
+- (bool)moveFileToTemporaryDirectoryFromSourcePath:(NSString *)sourcePath destinationPath:(NSString *)destinationPath {
+    TalkAccount *activeAccount = [[NCDatabaseManager sharedInstance] activeAccount];
+
+    [[NCAPIController sharedInstance] setupNCCommunicationForAccount:activeAccount];
```
SystemKeeper: We don't need to setup the communication here, as we only move locally.

NextcloudTalk/BaseChatViewController.swift
SystemKeeper: You can just use "file" as the key here

SystemKeeper:
Hey Daniel, left you a message about this PR on our instance :)

NextcloudTalk/BaseChatViewController.swift
DanielStandfest: added a check

DanielStandfest:
Review implemented.

Ivansss (approved):
Thanks a lot for this PR @DanielStandfest :) 
Works nicely now and it's a great enhancement!

SystemKeeper (approved):

SystemKeeper:
Thanks again @DanielStandfest ! Merged and will be included in the next release :)
Please remember to rebase instead of creating merge request though ;)

PR #1891, opened 2024-11-22
Fix fetching all locally stored messages for a conversation

SystemKeeper:
* Fixes regression from https://github.com/nextcloud/talk-ios/pull/1886

We did not enforce a message limit in call cases, regardless of a unread marker is involved or not. That lead to the problem, that we load all locally known messages of a conversation (to be more specific, all messages in the last known chat block).

In this PR, we
* Add the option to make sure, that a specific messageId is included in the batch (this is useful for the readmarker case)
* When we ensure a messageId, we try to load more messages, up to the limit (so _some_ history is already included in the readmarker case)
* We only count visible messages towards the limit, update messages will be ignored.

### DIFF
--- a/NextcloudTalk/NCChatController.m
+++ b/NextcloudTalk/NCChatController.m
@@ -71,7 +71,7 @@ NSString * const NCChatControllerDidReceiveMessagesInBackgroundNotification     
     return sortedBlocks;
 }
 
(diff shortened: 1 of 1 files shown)


Ivansss (approved):
Tested and works 👍

PR #1898, opened 2024-11-29
fix(UI): Perform badge view resizing and setting badge color without animations

Ivansss:


SystemKeeper (approved):
Tested and works - looks consistent now 👍

PR #1905, opened 2024-12-02
Render performance

SystemKeeper:
I noticed a few issues when scrolling and found some places where we can improve the time needed for each cell. Especially the invisible cells still needed quite some computing, even though they are not displayed:

CPU cycles before:
Bildschirmfoto 2024-11-30 um 17 15 55

After:
Bildschirmfoto 2024-11-30 um 17 18 56

So this PR:
* Does not include invisible messages in the table view
* Caches parsed markdown
* Caches auth parameters (before we queried the keychain for each avatar/message)
* Simplifies cell prepareForReuse
* Pass down TalkAccount (so we don't query it for each cell multiple times). Since we indirectly require a TalkAccount for the chat view controllers, we can also make it mandatory and pass it down.


SystemKeeper:
I saw (very very few) animation issues with this PR, nothing I would consider serious, but still too much to not see a coincidence with this PR. https://github.com/nextcloud/talk-ios/pull/1905/commits/5c17fd38a7b9692ca54547b3514c8ddd089a9292 reverts part of the reuse simplification, which I suspect could cause that.

Ivansss (approved):
Really nice improvement 👌

PR #1914, opened 2024-12-09
feat(PushNotifications): Allow to test push notifications

Ivansss:
Added option in diagnostics view to test push notifications:

IMG_E0101

Show results after the test and allow to copy them:

IMG_0100


SystemKeeper (commented):
Tested and works nicely.

One thing I noticed, there's no indication that the test request is running. So when the server is a bit slow, or the internet connection a bit weak, you might press the test button multiple times. Should we disable the test button (maybe with an activity indicator) until we received an API response?

Ivansss:
Added activity indicator and disable user interaction in cell while test is running

SystemKeeper (approved):

PR #1916, opened 2024-12-12
Added ability to view context of messages in media viewer

### LINKED ISSUE #1704: Add the ability to open message context in media view
When swiping through images in the media view, it is not possible to jump to the message that belongs to the image. We could add a way to show the message in the `ContextChatViewController` (much like it is done in the shared items dialog) to get some information about the message. Jumping to the message might not be possible, since it depends on if the message was loaded or not.

rapterjet2004:
- fixes #1704

### DIFF
--- a/NextcloudTalk/NCMediaViewerViewController.swift
+++ b/NextcloudTalk/NCMediaViewerViewController.swift
@@ -50,14 +50,16 @@ import UIKit
             if let account = message.account, let chatViewController = ContextChatViewController(forRoom: self.room, withAccount: account, withMessage: [], withHighlightId: 0) {
 
                 // Fetch the context of the message and update the BaseChatViewController
-                NCChatController(for: self.room).getMessageContext(forMessageId: message.messageId, withLimit: 10) { messages in
+                NCChatController(for: self.room).getMessageContext(forMessageId: message.messageId, withLimit: 50) { messages in
                     guard let messages else { return }
 
                     chatViewController.appendMessages(messages: messages)
                     chatViewController.reloadDataAndHighlightMessage(messageId: message.messageId)
+
                 }
 
-                self.present(chatViewController, animated: true)
+                let navController = NCNavigationController(rootViewController: chatViewController)
+                self.present(navController, animated: true)
             }
 
         })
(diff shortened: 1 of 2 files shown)


NextcloudTalk/NCMediaViewerViewController.swift
```diff
@@ -184,6 +212,7 @@ import UIKit
         self.navigationItem.title = mediaPageViewController.navigationItem.title
 
         self.shareButton.isEnabled = (mediaPageViewController.currentImage != nil) || (mediaPageViewController.currentVideoURL != nil)
+        self.showMessageButton.isEnabled = (mediaPageViewController.currentImage != nil) || (mediaPageViewController.currentVideoURL != nil) // TODO include check if msg was loaded
```
SystemKeeper: Since you fetch the context, we don't need to check if the message was loaded (I assume you. mean in the database)

SystemKeeper: This only works after a swipe, not when you open the media view for the first time.

NextcloudTalk/NCMediaViewerViewController.swift
```diff
@@ -38,7 +39,34 @@ import UIKit
...
+
+                // Fetch the context of the message and update the BaseChatViewController
+                NCChatController(for: self.room).getMessageContext(forMessageId: message.messageId, withLimit: 10) { messages in
+                    guard let messages else { return }
+
+                    chatViewController.appendMessages(messages: messages)
+                    chatViewController.reloadDataAndHighlightMessage(messageId: message.messageId)
+                }
+
+                self.present(chatViewController, animated: true)
```
SystemKeeper: Need's to be embedded in `NCNavigationController`

NextcloudTalk/NCMediaViewerViewController.swift
```diff
@@ -38,7 +39,34 @@ import UIKit
...
+        showMessageButton.isEnabled = false
+        showMessageButton.primaryAction = UIAction(title: "", image: .init(systemName: "text.magnifyingglass"), handler: { [unowned self, unowned showMessageButton] _ in
+            guard let mediaPageViewController = self.getCurrentPageViewController() else { return }
+
+            let message = mediaPageViewController.message
+
+            if let account = message.account, let chatViewController = ContextChatViewController(forRoom: self.room, withAccount: account, withMessage: [], withHighlightId: 0) {
+
+                // Fetch the context of the message and update the BaseChatViewController
+                NCChatController(for: self.room).getMessageContext(forMessageId: message.messageId, withLimit: 10) { messages in
```
SystemKeeper: In shared items list we use a limit of 50, I think that would be fine here too.

NextcloudTalk/NCMediaViewerViewController.swift
```diff
@@ -91,7 +119,7 @@ import UIKit
         self.navigationController?.toolbar.compactAppearance = appearance
         self.navigationController?.toolbar.scrollEdgeAppearance = appearance
 
-        self.toolbarItems = [shareButton]
+        self.toolbarItems = [shareButton, showMessageButton]
```
SystemKeeper: I would try to add a bit of padding here.

SystemKeeper (commented):
Very quick review comments ;)

SystemKeeper (approved):
Nice, thank you!
I noticed a few small things when testing this (not only related to this PR), we can fix in a follow-up.

PR #1919, opened 2024-12-15
fix(textView): Disable adaptive glyph support

SystemKeeper:
Otherwise we can add memojis and genmojis to the textView without handling them correctly.


Ivansss:
Closing in favour of #1920

PR #1936, opened 2025-01-10
fix(api): Use request account instead of app's active account

Ivansss:


NextcloudTalk/NCAPIController.m
```diff
@@ -324,8 +324,7 @@ - (NSURLSessionDataTask *)getContactsForAccount:(TalkAccount *)account forRoom:(
         NSMutableArray *users = [[NSMutableArray alloc] initWithCapacity:responseContacts.count];
         for (NSDictionary *user in responseContacts) {
             NCUser *ncUser = [NCUser userWithDictionary:user];
-            TalkAccount *activeAccount = [[NCDatabaseManager sharedInstance] activeAccount];
-            if (ncUser && !([ncUser.userId isEqualToString:activeAccount.userId] && [ncUser.source isEqualToString:kParticipantTypeUser])) {
+            if (ncUser && ([ncUser.userId isEqualToString:account.userId] && [ncUser.source isEqualToString:kParticipantTypeUser])) {
```
SystemKeeper: Hm... This changes the condition?

SystemKeeper (changes requested):

NextcloudTalk/NCAPIController.m
Ivansss: Oopsie, good catch! push that change by mistake 😅

SystemKeeper (approved):

PR #1945, opened 2025-01-16
feat(conversations): Support conversation description length capability

### LINKED ISSUE #1940: Support conversation description length from capabilities
Hardcoded value of 500 in the past has to be replaced with a capability check for:
> `config => conversations => description-length` (local) - The maximum length for conversation descriptions, currently 2000. Before this config was added the implicit limit was 500, since the existance of the feature capability `room-description`.

the 500 still need to be kept as a fall back for older server versions

- Changed in https://github.com/nextcloud/spreed/pull/14124

Ivansss:
Fix #1940


SystemKeeper:
~~Empty poll drafts is missing a close button~~

Sorry, wrong PR ...

SystemKeeper (approved):

PR #1952, opened 2025-01-21
perf: Read number of pending invitations from header

SystemKeeper:
* With https://github.com/nextcloud/spreed/pull/11944 we can read the pending invites from the room request and don't need to separately query the invites endpoint.

~~* Might even make more sense with https://github.com/nextcloud/spreed/pull/14168, if we can add that.~~

If the header is not there, we don't have pending invitations.

### DIFF
--- a/NextcloudTalk/NCDatabaseManager.m
+++ b/NextcloudTalk/NCDatabaseManager.m
@@ -16,7 +16,7 @@
 
 NSString *const kTalkDatabaseFolder                 = @"Library/Application Support/Talk";
 NSString *const kTalkDatabaseFileName               = @"talk.realm";
-uint64_t const kTalkDatabaseSchemaVersion           = 74;
+uint64_t const kTalkDatabaseSchemaVersion           = 75;
 
 NSString * const kCapabilitySystemMessages          = @"system-messages";
 NSString * const kCapabilityNotificationLevels      = @"notification-levels";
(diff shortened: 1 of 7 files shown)


NextcloudTalk/NCDatabaseManager.m
```diff
@@ -16,7 +16,7 @@
 
 NSString *const kTalkDatabaseFolder                 = @"Library/Application Support/Talk";
 NSString *const kTalkDatabaseFileName               = @"talk.realm";
-uint64_t const kTalkDatabaseSchemaVersion           = 73;
+uint64_t const kTalkDatabaseSchemaVersion           = 74;
```
Ivansss: I guess the idea behind bumping the schema version was to remove `lastPendingFederationInvitationFetch` because it is no longer used. Forgot to push that change? :) 
https://github.com/nextcloud/talk-ios/blob/0a2995c43ac390308339a7e041f299668625ec1e/NextcloudTalk/TalkAccount.h#L44

Ivansss (changes requested):

NextcloudTalk/NCDatabaseManager.m
SystemKeeper: Good catch! Thanks, indeed!

Ivansss (approved):

PR #1965, opened 2025-01-30
feat(polls): Allow editing and deleting poll drafts

Ivansss:
IMG_E0129


SystemKeeper (approved):
Tested and works. The only thing I find a bit confusing is that "Cancel" on edit will close the whole controller and it is not like "going back to creation". But I don't think that's a big issue.

PR #1968, opened 2025-02-04
feat(meetings): Show upcoming events 📆 

Ivansss:
IMG_E0131

### DIFF
--- a/NextcloudTalk/CalendarEvent.swift
+++ b/NextcloudTalk/CalendarEvent.swift
@@ -5,15 +5,15 @@
 
 import Foundation
 
-@objcMembers public class CalendarEvent: NSObject {
+struct CalendarEvent {
 
-    public var calendarAppUrl: String
-    public var calendarUri: String
-    public var location: String
-    public var recurrenceId: String
-    public var start: Int
-    public var summary: String
-    public var uri: String
+    var calendarAppUrl: String
+    var calendarUri: String
+    var location: String
+    var recurrenceId: String
+    var start: Int
+    var summary: String
+    var uri: String
 
     init(dictionary: [String: Any]) {
         self.calendarAppUrl = dictionary["calendarAppUrl"] as? String ?? ""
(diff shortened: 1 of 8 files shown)


NextcloudTalk/CalendarEvent.swift
```diff
@@ -0,0 +1,29 @@
+//
+// SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+// SPDX-License-Identifier: GPL-3.0-or-later
+//
+
+import Foundation
+
+@objcMembers public class CalendarEvent: NSObject {
```
SystemKeeper:
```suggestion
public class CalendarEvent {
```

If only used from swift, we can remove the objc interoperability (need to mark the api method with `@nonobjc`). In theory can also be a  `struct` then.

NextcloudTalk/CalendarEvent.swift
```diff
@@ -0,0 +1,29 @@
...
+    init(dictionary: [String: Any]) {
+        self.calendarAppUrl = dictionary["calendarAppUrl"] as? String ?? ""
+        self.calendarUri = dictionary["calendarUri"] as? String ?? ""
+        self.location = dictionary["location"] as? String ?? ""
+        self.recurrenceId = dictionary["recurrenceId"] as? String ?? ""
+        self.start = dictionary["start"] as? Int ?? -1
+        self.summary = dictionary["summary"] as? String ?? ""
+        self.uri = dictionary["uri"] as? String ?? ""
+
+        super.init()
```
SystemKeeper:
```suggestion
```

Not needed when not a subclass of NSObject

NextcloudTalk/NCAPIControllerExtensions.swift
```diff
@@ -680,4 +680,26 @@ import Foundation
             completionBlock(message)
         }
     }
+
+    // MARK: - Upcoming events
+
+    public func upcomingEvents(_ room: NCRoom, forAccount account: TalkAccount, completionBlock: @escaping (_ events: [CalendarEvent]) -> Void) {
```
SystemKeeper: Add `@nonobjc` here in case you wanna remove the objc part.

NextcloudTalk/NCUtils.swift
```diff
@@ -247,6 +247,52 @@ import AVFoundation
         return Calendar.current.date(byAdding: .day, value: (weekday - currentWeekday), to: date)!
     }
 
+    public static func eventTime(from timestamp: TimeInterval) -> String {
```
SystemKeeper: Personally I would move that to CalendarEvent instead of having it as a static in NCUtils?

SystemKeeper (commented):
I am not sure about the `eventTime` method. This seems problematic, translation-wise. shouldn't we better use DateFormatter with the desired format and enable relative formatting and not handle that ourselves?

Ivansss:
Couldn't find a way to display event `start` time the same way as in the web using `RelativeDateTimeFormatter()`

SystemKeeper:
For today and tomorrow I would go with

```swift
let dateFormatter = DateFormatter()
dateFormatter.dateStyle = .long
dateFormatter.timeStyle = .short
dateFormatter.doesRelativeDateFormatting = true

let now = Date()
let tomorrow = Calendar.current.date(byAdding: .day, value: 1, to: now)!

print(dateFormatter.string(from: now))
print(dateFormatter.string(from: tomorrow))
```

which prints for my locale:

```
Heute um 20:49
Morgen um 20:49
```

Personally I would also be fine to just use it for all other cases as well, it would show:

```
12. Februar 2025 um 20:49
```

for next week.

`.full` would also work for `dateStyle` showing:

```
Mittwoch, 12. Februar 2025 um 20:51
```

Alternatively (if we can live without the "at"):

```swift
let localeArray = ["en_US", "de_DE"]
for localeID in localeArray {
    print(now.formatted(.dateTime
             .weekday(.wide)
             .hour(.conversationalTwoDigits(amPM: .wide))
             .minute(.defaultDigits)
             .locale(Locale(identifier: localeID))
    ))
}
```

could be used for the < 1 week and displays:

```
Wednesday 08:55 PM
Mittwoch, 20:55
```

SystemKeeper (approved):
👌

PR #1993, opened 2025-02-19
feat(meeting): 📆 Allow to schedule meetings

Ivansss:


NextcloudTalk/NCAPIController+CalDAV.swift
```diff
@@ -0,0 +1,107 @@
+//
+// SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+// SPDX-License-Identifier: GPL-3.0-or-later
+//
+
+import Foundation
+import SwiftyXMLParser
```
SystemKeeper, line 7: `SwiftyXMLParser` is only a sub-dependency of another package, that's why "build for testing" fails.

SystemKeeper (approved):

PR #1995, opened 2025-02-20
fix(login): Present login screen after token revoked (and no other accounts configured)

Ivansss:


SystemKeeper (approved):
Nice, this bugged me for a while :)

PR #2007, opened 2025-03-02
Add keyboard support to autocompletion

### LINKED ISSUE #1999: Keyboard Interactive Mention Suggestions
I use my iPad with an external keyboard. When I start to type a mention, I get the usual list of suggestions. I cannot navigate or select the suggestions by keyboard, though. I have to tap on the screen. If I hit return while having an incomplete mention, that fragment is posted as a message.

As natural in desktop environments, the list items should be focusable and navigable by arrow keys of external periphery. Hitting return should selected the highlighted suggestion. This would also increase accessibility.

Tapping on the screen is the default and only alternative and exactly what I want to avoid by using an external keyboard with physical keys. 😁

**Additional context**
None.

SystemKeeper:
* Fixes https://github.com/nextcloud/talk-ios/issues/1999


i2h3:
I tried this out with my physical iPad and Magic Keyboard (via integration with macOS) and it works fine for me.

i2h3 (approved):

Ivansss (approved):

PR #2083, opened 2025-05-08
chore(i18n): Add comments for translators

rakekniven:
Reported at Transifex


SystemKeeper:
@rakekniven Adjusted your changes in a second PR to make the difference clear. You can either change both locations (the localizable file and the source files), or (recommended when you use a Mac and cloned the repo) adjust the source and then run `generate-localizable-strings-file.sh` in the root of this repo. That is basically what CI does not check if there are changes to localization that are not correctly reflected in the localizable file.

Changes okay this way?

SystemKeeper (approved):

rakekniven (commented):
Fine for me

PR #2084, opened 2025-05-08
chore(i18n): Added comment for translators

rakekniven:
Reported at Transifex

### DIFF
--- a/NextcloudTalk/DiagnosticsTableViewController.swift
+++ b/NextcloudTalk/DiagnosticsTableViewController.swift
@@ -233,7 +233,7 @@ class DiagnosticsTableViewController: UITableViewController {
             return NSLocalizedString("Server", comment: "")
 
         case DiagnosticsSections.kDiagnosticsSectionTalk.rawValue:
-            return NSLocalizedString("Talk", comment: "")
+            return "Talk"
 
         case DiagnosticsSections.kDiagnosticsSectionSignaling.rawValue:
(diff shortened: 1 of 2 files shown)


rakekniven:
@SystemKeeper Will my adaption work here?

SystemKeeper:
> @SystemKeeper Will my adaption work here?

I think in this case the easiest thing to do is to just remove the translation, since we don't want to translate it anyway. I changed your commit accordingly :)

SystemKeeper (approved):

rakekniven:
> > @SystemKeeper Will my adaption work here?
> 
> I think in this case the easiest thing to do is to just remove the translation, since we don't want to translate it anyway. I changed your commit accordingly :)

Yeah, you could have told me that in the chat :) 

Cheers

SystemKeeper:
> Yeah, you could have told me that in the chat :)

Actually no, at least not at the time we talked about this. I first needed to clarify if using "Talk" is fine here or if we need to use the product name, as defined at:

https://github.com/nextcloud/talk-ios/blob/3d8f3b351f904595dffd3a2757bb3acc4bb5a7b9/NextcloudTalk/NCAppBranding.m#L21

My comment in the chat had the reason that you know that it works a bit differently in this repo (and since you opened 2 PRs), which will make CI fail and that you don't have additional work figuring out what's wrong.

Your work is really appreciated!

PR #2088, opened 2025-05-15
Clearer Mark as Read/Unread icons

kra-mo:
These are more consistent with other apps and thus clearer to users. We can't use bubble as there is no badge symbol for it unfortunately.

### DIFF
--- a/NextcloudTalk/RoomsTableViewController.m
+++ b/NextcloudTalk/RoomsTableViewController.m
@@ -1410,7 +1410,7 @@ typedef enum RoomsSections {
             completionHandler(true);
         }];
 
-        NSString *markImageName = (room.unreadMessages > 0) ? @"eye" : @"eye.slash";
+        NSString *markImageName = (room.unreadMessages > 0) ? @"app.badge.checkmark" : @"app.badge";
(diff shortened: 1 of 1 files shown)


Ivansss:
Hi @kra-mo :)

Thank you so much for the PR!

This is how the read/unread icons look in the web:
| Read | Unread |
|--------|------|
| Screenshot 2025-05-15 at 16 48 43 | Screenshot 2025-05-15 at 16 48 12 |

We always try to use the same or similar icons as in the web and android app to keep consistency.
I guess we should look for one icon that works well across all platforms (web/desktop, android and iOS) and make the change on all of them at the same time.

kra-mo:
I see, for Material, `mark_unread_chat_alt` and `mark_chat_read` are similar and more semantic as well.

What would be the next steps for changing these?

SystemKeeper:
Hello @kra-mo,

we had some exchange with @nimishavijay this week about this. We still want to try to use SF icons on iOS if possible somehow, but the proposed icons might not be a good fit right now. We discussed using something from the `bubble.*` range of icons maybe, but @nimishavijay will reach to you on that one :) 
For now I'll close the PR, to not have it around open for too long. Please reopen it, when there was a conclusion reached :-)

PR #2090, opened 2025-05-15
fix(call): Send notifications to newly added participants

### LINKED ISSUE #2087: Send silentFor array parameter when switch-to an extended_conversation
- To resolve https://github.com/nextcloud/spreed/issues/14983 clients need to tell the backend to send call notifications and ignore the previous participant
- API in progress in https://github.com/nextcloud/spreed/pull/15083
- Starting point should be
  https://github.com/nextcloud/talk-android/blob/1e1a8f7f21c0b3d2f74a4278bb3f09703b0e0533/app/src/main/java/com/nextcloud/talk/signaling/SignalingMessageReceiver.java#L295
- API call where you need to send the silentFor is
  https://github.com/nextcloud/talk-android/blob/1e1a8f7f21c0b3d2f74a4278bb3f09703b0e0533/app/src/main/java/com/nextcloud/talk/chat/ChatActivity.kt#L3024

The silentFor array can be set to the conversation.name of the previous conversation

Ivansss:
Fix #2087

### DIFF
--- a/NextcloudTalk/CallViewController.swift
+++ b/NextcloudTalk/CallViewController.swift
@@ -128,6 +128,7 @@ class CallViewController: UIViewController,
     private var callDurationTimer: Timer?
     private var soundsPlayer: AVAudioPlayer?
     private var currentCallState: CallState = .joining
+    private var previousParticipants: [String] = []
 
     public init?(for room: NCRoom, asUser displayName: String, audioOnly: Bool) {
         self.room = room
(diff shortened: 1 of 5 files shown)


SystemKeeper (approved):

NextcloudTalk/CallViewController.swift
```diff
@@ -852,9 +861,18 @@ class CallViewController: UIViewController,
...
+                self.previousParticipants.removeAll()
+                if self.room.type == .oneToOne {
+                    self.previousParticipants.append(self.room.name)
+                }
+
                 // Assign new room as current room
                 self.room = newRoom
 
+                // Start call silently in new room
+                self.silentCall = true
```
mahibi, line 874: @Ivansss @SystemKeeper 
From what i understand, `silentCall` should be `false` here?
Because want that the call is ringing for all new participants, except for the `previousParticipants`, right?

PR #2124, opened 2025-06-11
Chore/noid/remove afnetwork image

SystemKeeper:
Currently AFnetwork is used soley for file previews. For everything else (e.g. avatars) we use SDWebImage nowadays. The API works a bit differently, but I think it makes sense to only rely on one library here.


NextcloudTalk/CallViewController.swift
```diff
@@ -144,9 +144,6 @@ class CallViewController: UIViewController,
 
         self.modalPresentationStyle = .fullScreen
 
-        // Use image downloader without cache so I can get 200 or 201 from the avatar requests.
-        AvatarBackgroundImageView.setSharedImageDownloader(NCAPIController.sharedInstance().imageDownloaderNoCache)
```
SystemKeeper, line 148: Side note: Wasn't even used anymore...

Ivansss (approved):

PR #2129, opened 2025-06-16
feat(login): Login view improvements

Ivansss:
| Login view (Before) | Login view (After) |
| --- | --- |
| IMG_0148 | IMG_0155 |

| QR scan view (Before) | QR scan view (After) |
| --- | --- |
| IMG_0152 | IMG_0154 |


nimishavijay:
Looks way better! :) Definitely more clear with the outline for the server address field :) One question: would it be possible to show the hint on where to find the QR code on the screen after you click on the QR code button? Something like this:

image

Otherwise we can create an intermediate screen when you click on the QR button with just the steps (similar to Whatsapp's link device screen) What do you think?

nimishavijay:
Nice! The buttons look more iOS-y now :) Only some small notes:

- [x] I don't use iOS that much but I think making the border width 1px again, and increasing the border radius by around 2px or 4px will make it look more native (what do you think?)
- [x] The close button on the QR code screen can be on top-left corner instead of top-right as usually that is where a positive action is located

Ivansss:
Adjusted and updated screenshots above :) @nimishavijay

NextcloudTalk/LoginViewController.swift
```diff
@@ -0,0 +1,334 @@
...
+              !serverTextFieldText.isEmpty else {
+            serverTextField.becomeFirstResponder()
+            return
+        }
+
+        startLoginProcess(serverURL: serverTextFieldText, user: nil)
+    }
+
+    func startLoginProcess(serverURL: String, user: String?) {
+        var normalizedServerURL = serverURL
```
SystemKeeper:
```suggestion
        var normalizedServerURL = serverURL.lowercased()
```

to make sure the next check with `hasPrefix` works in all cases?

NextcloudTalk/LoginViewController.swift
```diff
@@ -0,0 +1,334 @@
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
```
SystemKeeper:
```suggestion
//
// SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
// SPDX-License-Identifier: GPL-3.0-or-later
//
```

To make swiftlint happy

NextcloudTalk/LoginViewController.swift
```diff
@@ -0,0 +1,334 @@
...
+        // Import account button
+        importAccountButton.setTitle(NSLocalizedString("Import account", comment: ""), for: .normal)
+
+        // Buttons style
+        [loginButton, qrCodeButton, importAccountButton].forEach { button in
+            button.backgroundColor = NCAppBranding.brandColor()
+            button.layer.borderColor = NCAppBranding.brandTextColor().cgColor
+            button.layer.cornerRadius = 12
+            button.layer.borderWidth = 1
+            button.clipsToBounds = true
```
SystemKeeper, line 89: Just wondering, would it make sense / be possible to re-use `NCButton` here?

NextcloudTalk/QRScannerViewController.swift
```diff
@@ -0,0 +1,166 @@
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
```
SystemKeeper:
```suggestion
//
// SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
// SPDX-License-Identifier: GPL-3.0-or-later
//
```

SystemKeeper (commented):
It's not an issue of this PR, but I am wondering, if we should have all 3 buttons below each other. While the first one belongs to the text field, and the others have a distinct action. Should there be a separator or a bit more space between "Log in" and "Scan QR code"?

nimishavijay:
@SystemKeeper You mean like this? Yes, definitely makes sense :) 

image

NextcloudTalk/LoginViewController.swift
Ivansss: NCButton has `cornerRadius = 8`, I would just leave this for now

Ivansss:
@nimishavijay @SystemKeeper the only problem I see is that `Scan QR code` and `Import account` might not always be there.
When there is only one option, as in the example below, the button appears misaligned and looks out of place.

| All options | One option |
| --- | --- |
| IMG_0170 | IMG_0171 |

SystemKeeper:
Would still be fine by me, since it’s still a destinct action. @nimishavijay any thoughts ?

SystemKeeper (approved):

PR #2140, opened 2025-06-27
fix: Avoid app crash when removing account and fetching profile image takes some time

Ivansss:


SystemKeeper (changes requested):
I am not sure why this crashes. Does this happen when the view controller is not visible anymore, and we try to reload the section? Or is the section somehow invalid?
We changed this part a while back to not reload the whole table view to prevent hitches in the settings controller while scrolling.

I feel like I am missing something here

SystemKeeper:
As discussed, since we can't reproduce right now -> close

PR #2141, opened 2025-07-02
chore(chat): Add new date headers

Ivansss:
- [x] Dynamic font
- [x] Flexible width
- [x] Swift

| Before | After |
| --- | --- |
| IMG_E0159 | IMG_E0158 |


SystemKeeper (approved):
Nice work!

PR #2149, opened 2025-07-09
🧵 Threads

Ivansss:
Missing features:

- [x] Show rooms thread (either creating a more-options menu where the calendar button is placed or in room info view)
- [x] Show thread title in thread view
- [ ] Show number of replies (original thread message, thread view subtitle)
- [x] Allow to set notification level inside thread view (top right button in navigation bar)
- [ ] Allow to send attachments directly to a thread


NextcloudTalk/NCAPIControllerExtensions.swift
```diff
@@ -1156,4 +1156,73 @@ import NextcloudKit
             completionBlock(ProfileInfo(dictionary: dataDict))
         }
     }
+
+    // MARK: - Threads
+
+    @nonobjc
+    public func getThreads(for accountId: String, in roomToken: String, offsetId: Int? = 0, completionBlock: @escaping (_ threads: [ThreadInfo]?) -> Void) {
```
SystemKeeper: Feels like `offsetId` does not need to be an optional, if we check for `> 0` anyway?

NextcloudTalk/ThreadInfo.swift
```diff
@@ -0,0 +1,28 @@
...
+    var thread: MessageThread
+    var attendee: ThreadAttendee
+    var first: NCChatMessage
+    var last: NCChatMessage
+
+    init(thread: MessageThread, attendee: ThreadAttendee, first: NCChatMessage, last: NCChatMessage) {
+        self.thread = thread
+        self.attendee = attendee
+        self.first = first
+        self.last = last
```
SystemKeeper:
```suggestion
        self.firstMessage = firstMessage
        self.lastMessage = lastMessage
```

?

NextcloudTalk/ThreadInfo.swift
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    init(dictionary: [String: Any]) {
+        self.thread = MessageThread(dictionary: dictionary["thread"] as? [String: Any] ?? [:])
+        self.attendee = ThreadAttendee(dictionary: dictionary["attendee"] as? [String: Any] ?? [:])
+        self.first = NCChatMessage(dictionary: dictionary["first"] as? [String: Any] ?? [:])
+        self.last = NCChatMessage(dictionary: dictionary["last"] as? [String: Any] ?? [:])
```
SystemKeeper: Just a feeling, maybe we should go with optionals here?

NextcloudTalk/NCAPIControllerExtensions.swift
Ivansss: The `/threads/recent` endpoint doesn't have an `offsetId` so removing it

NextcloudTalk/ThreadInfo.swift
Ivansss: makes it more readable 👍

NextcloudTalk/ThreadInfo.swift
Ivansss: `last` could be optional, I will change it ;)

NextcloudTalk/ChatViewController.swift
```diff
@@ -155,6 +159,22 @@ import SwiftUI
...
+    private lazy var closeButton: UIBarButtonItem = {
+        let closeButton = UIBarButtonItem(title: nil, style: .plain, target: nil, action: nil)
+
+        closeButton.primaryAction = UIAction(title: NSLocalizedString("Close", comment: ""), handler: { _ in
+            if self.presentedInCall {
+                NCRoomsManager.sharedInstance().callViewController?.toggleChatView()
+            } else {
+                self.dismiss(animated: true)
+            }
+        })
```
SystemKeeper: `self` is captured here...

SystemKeeper (approved):

PR #2153, opened 2025-07-28
Chat redesign

SystemKeeper:
- [x] Adjust tests
- [x] Fix padding issue when the app user quotes a message
- [x] Check on cell flickering when adding/removing a reaction
- [x] Check system message width calculation
- [x] Context menu mask for grouped messages is not correct
- [x] Adjust expanded system messages background 
- [ ] On edit, quote is missing for a split second (seems unrelated to this PR)
- [x] Adjust background Color of syntax blocks 
- [x] OOO Background adjustment
- [x] File sharing background


rpinto73:
Thank you!

PR #2160, opened 2025-08-16
fix: Use downscaled images to prevent memory issues

### LINKED ISSUE #2157: SpringBoard crash when viewing high resolution PNG files
## Steps to reproduce
1. Send a high resolution PNG file in a chat. (In my example it was a 19268 × 25022 pixels image with 2.3 MB)
2. Try to view that image by tapping on in the chat.

### Expected behaviour
The image preview screen slides in from the bottom, loads the image and shows it. Also a small preview of the image should be visible in the chat.

### Actual behaviour
A full SpringBoard crash happens: The phone shows a black screen with a white loading circle and it returns to the Lock-Screen.

## Device information

**Device:** iPhone 12 Pro  AND  iPhone 13 Mini

**iOS version:** 18.6  AND  18.5

**Talk version:** 21.1.1

## Server information
Unfortunately, I'm not the administrator, so I don't have access to all of that information. However, for this specific bug, it shouldn't be relevant. If it unexpectedly turns out to be important, I can provide that information.

### Attachement
The image that causes the crash:
Image

SystemKeeper:
* Fixes https://github.com/nextcloud/talk-ios/issues/2157
* 
Simulator Screenshot - iPhone 16 Pro - 2025-08-16 at 19 39 04

### DIFF
--- a/NextcloudTalk/NCMediaViewerPageViewController.swift
+++ b/NextcloudTalk/NCMediaViewerPageViewController.swift
@@ -74,6 +74,37 @@ import SwiftyGif
         return errorView
     }()
 
+    private lazy var downscaledWarningView = {
+        let warningView = UIView()
+        warningView.translatesAutoresizingMaskIntoConstraints = false
+
+        let warningImage = UIImageView()
+        warningImage.image = UIImage(systemName: "exclamationmark.triangle.fill")
+        warningImage.contentMode = .scaleAspectFit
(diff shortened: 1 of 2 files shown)


SystemKeeper:
Going to 2048 as some devices in test still crashed with larger images (not sure if it's the image or the downscaling, though)

Ivansss (approved):

PR #2175, opened 2025-09-02
Fixed the UI not updating when the room is Locked/Unlocked

### LINKED ISSUE #2082: Hide/show inputbar when conversation is locked/unlocked
Not sure if this is a regression or if it was always like that, but when a conversation is locked from a second device, the first does not adapt the inputbar accordingly.

rapterjet2004:
- Based on https://github.com/nextcloud/talk-ios/pull/2568 to make it work with chat relay.
- fixes #2082 

Updated the room upon receiving a `processRoomListEvent` in the `NCExternalSignalingController`.

Is this the right way to implement this? Not sure if updating is a heavy operation or not. On [Android](https://github.com/nextcloud/talk-android/pull/4567/files#diff-a3567d71e5bb91cd3a5b0f732565ce89ebb3edcaa51f45dfe36922cb1b4e7259R679) we have the room state updated upon receiving a new capability from the server, which is then reflected in the UI.

https://github.com/user-attachments/assets/a6f4cf8b-38e5-4364-9d7a-c2a0bf317b41

### DIFF
--- a/NextcloudTalk/Chat/ChatViewController.swift
+++ b/NextcloudTalk/Chat/ChatViewController.swift
@@ -550,6 +550,7 @@ import SwiftUI
         NotificationCenter.default.addObserver(self, selector: #selector(didReceiveThreadMessage(notification:)), name: NSNotification.Name.NCChatControllerDidReceiveThreadMessage, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(didReceiveThreadNotFound(notification:)), name: NSNotification.Name.NCChatControllerDidReceiveThreadNotFound, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(didReceiveHistoryCleared(notification:)), name: NSNotification.Name.NCChatControllerDidReceiveHistoryCleared, object: nil)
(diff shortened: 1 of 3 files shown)


SystemKeeper:
Cool stuff, thanks! Need to take a closer look of that is the best way to solve, out of my head I am not totally sure.

Ivansss:
Do you also re-fetch the room from the server whenever you receive an `update` message from the external signaling on Android?
Even though that solution works, I think the best and most performant solution would be to check the updated attributes included in the `update` message and update the room locally, without making an additional request to the server.

SystemKeeper:
The recently merged pinning PR is doing something similar to directly act on the system message:

https://github.com/nextcloud/talk-ios/blob/ee12fd34c24b95dd67b867606a69e7ede05cf855/NextcloudTalk/Chat/ChatViewController.swift#L1785-L1792

I think we can adopt this for this here as well?

SystemKeeper:
@rapterjet2004 You plan on working on this any time soon? We can put it to draft or take over?

SystemKeeper:
@rapterjet2004 🏓

SystemKeeper (approved):
Tested and works nicely.

PR #2179, opened 2025-09-09
fix: Thread chat controller

Ivansss:


NextcloudTalk/NCChatController.m
```diff
@@ -702,23 +656,117 @@ - (void)getHistoryBatchFromMessagesId:(NSInteger)messageId
...
+
+        // Error handling
+        if (error) {
+            completion(nil, 0, error, statusCode);
+            return;
+        }
+
+        // Store new messages
+        if (forInitialChatHistory && isFirstIteration) {
+            [self updateChatBlocksWithReceivedMessages:messages newestKnown:messageId andLastKnown:lastKnownMessage];
```
SystemKeeper: Can we comment on why we need to do that here?

NextcloudTalk/NCChatController.m
```diff
@@ -260,7 +260,9 @@ - (void)updateChatBlocksWithLastKnown:(NSInteger)lastKnown
     if (lastKnown <= 0) {
         return;
     }
-    
+
+    NSInteger oldestMessageKnown = _threadId > 0 && lastKnown < _threadId ? _threadId : lastKnown;
```
SystemKeeper: A small explanation might make sense?

NextcloudTalk/NCChatController.m
```diff
@@ -702,23 +656,117 @@ - (void)getHistoryBatchFromMessagesId:(NSInteger)messageId
...
+                    completion(history, lastCommonReadMessage, nil, 0);
+                    return;
+                }
+            }
+
+            // Recursively fetch messages until finding visible ones
+            [self fetchHistoryUntilVisibleFromMessageId:lastKnownMessage
+                                  forInitialChatHistory:forInitialChatHistory
+                                       isFirstIteration:NO
+                                             completion:completion];
```
SystemKeeper, line 777: Would that be an infinite recursion when we reached the end and did not get visible messages? (Status code 304)

Ivansss: As discussed in private, 304 is an error response that will stop the recursion.

NextcloudTalk/NCChatController.m
Ivansss: Comment added

NextcloudTalk/NCChatController.m
Ivansss: Comment added

SystemKeeper (approved):

PR #2182, opened 2025-09-14
feat: Show thread title and number of replies

Ivansss:


SystemKeeper:
https://github.com/user-attachments/assets/45431b18-4459-4fc3-8ad9-beb42dc0918a

Something is not fully right when the context menu cell is created, I guess it's due to different constraints. Haven't looked to closely, but fine by me to keep as a follow-up.

SystemKeeper (approved):

PR #2190, opened 2025-09-18
fix: Status stack view alignment

Ivansss:
Before:
IMG_E0226

Now:
IMG_E0225


SystemKeeper (approved):

SystemKeeper (approved):
Might indeed be a leftover from something that isn't an issue anymore. At least I can't see any issue right now with this.

PR #2241, opened 2025-10-19
fix: Update thread original message on thread created

SystemKeeper:
* Ref https://github.com/nextcloud/spreed/issues/16160

We have the update functionality already, but it is skipped for `thread_created` messages. I think it's a relict of the first thread implementation that we had. Not sure if I overlook something here?


SystemKeeper:
Compared the two messages (original message and the one in the parent), seems to be fine to update, IMO:

image

Ivansss (approved):

PR #2242, opened 2025-10-19
fix: Adjust background color usage on iOS 26

### LINKED ISSUE #2228: Incorrect background color on macOS
Whatever is chosen for the background color of the app itself seems to render incorrectly when running on macOS. The whole app, including the sidebar becomes a dull gray. This also makes the sidebar indistinguishable.

This did not happen before the latest update.

Image

Not sure if this is just a quirk of this color on the OS or whether something can be done about it but it looks rather ugly.

PS: Honestly, I would expect chats to have a white background color regardless even on iPhone/iPad, as do most chat apps, the issue is just more pronounced here.

## Device information

**macOS version:** 26.1

**Talk version:** 22.0.0

SystemKeeper:
* Fixes #2228 

- [x] Test changes on iOS < 26

image


SystemKeeper:
Pushed 2 small adjustments and tested iOS 18, seems to work fine so far.

Ivansss (approved):

PR #2308, opened 2025-12-09
fix: Adjust AirPlay button position in call view

Ivansss:
| Before | After |
| --- | --- |
| IMG_E0302 | IMG_E0303 |


SystemKeeper (approved):
Whoops.

PR #2325, opened 2025-12-21
chore: Replace openSSL with native methods

SystemKeeper:
Quick test to use Apples Framework (and SwiftyRSA as a wrapper) instead of OpenSSL. Some things are not ideal, but works in general.

- [ ] ~~CCCertificate has a different format (BEGIN/END is missing). So will retrigger a verification popup~~
- [x] Some methods in SwiftyRSA are not accessible anymore (internal vs public default)
- [x] Public/PrivateKey naming conflicts with private framework from Apple 
- [x] Double check if we need to regenerate the keys or if everything is working with "old" keys and "new" APIs

* Ref https://github.com/nextcloud-deps/SwiftyRSA/pull/1


SystemKeeper:
> CCCertificate has a different format (BEGIN/END is missing). So will retrigger a verification popup

Not sure if it is worth it to try to have the same format and keep that around. I'd vote for leaving that out, therefore marking this PR as ready to review.

Ivansss (approved):

PR #2330, opened 2025-12-30
OOO/Pinned message view label calculation

SystemKeeper:
Otherwise the label is not layed out yet and therefore we calculate the height with a wrong value. This leads to more lines being shown, than expected.
Same is possible in pinned message view.


SystemKeeper:
* Closed in favor of https://github.com/nextcloud/talk-ios/pull/2331

PR #2333, opened 2026-01-04
fix: Expire background update of rooms

SystemKeeper:
* Currently we do update and delete in 2 separate transactions -> do it in one, I don't see a reason to have two here.
* If the background job expired, rollback the transaction and stop the execution


SystemKeeper:
It was a single transcation before https://github.com/nextcloud/talk-ios/commit/9112c0e425f14cafc3960508b27fd80bfce6e761#diff-eba5e69a1812d1585a2db89ef47e4a8f93434c27a992790126e164b7e36aa170. I don't see why we shouldn't do it in one?

Ivansss (approved):

PR #2367, opened 2026-01-26
fix(chat): Preserve message text when sharing media from picker

luflow:
## Summary

- Fixes an issue where text typed in the chat input field was lost when selecting media to share
- The text is now transferred to the ShareConfirmationViewController as the media caption
- Matches the existing behavior for pasted media content

## Problem

When a user had already typed text in the chat input field and then selected an image, video, or document to share via the picker, the text was being wiped/cleared instead of being used as the caption for the shared media.

## Solution

Added calls to `setChatMessage()` in the picker delegate methods to transfer the existing text to the ShareConfirmationViewController before clearing the main input field. This mirrors the implementation that already exists in `didPasteMediaContent()`.

**Affected picker delegates:**
- `PHPickerViewControllerDelegate` (photo library)
- `UIImagePickerControllerDelegate` (camera)
- `UIDocumentPickerViewControllerDelegate` (files)

If the user cancels the share, the text is restored to the chat input (this was already implemented in `shareConfirmationViewControllerDidCancel`).

## Test plan

- [ ] Type some text in the chat input
- [ ] Tap the attachment button and select "Photo Library"
- [ ] Select an image
- [ ] Verify the text appears in the caption field of the share confirmation view
- [ ] Cancel the share and verify the text is restored to the chat input
- [ ] Repeat for camera and file picker

### DIFF
--- a/NextcloudTalk/Chat/BaseChatViewController.swift
+++ b/NextcloudTalk/Chat/BaseChatViewController.swift
@@ -1599,6 +1599,9 @@ import SwiftUI
 
         guard let (shareConfirmationVC, navigationController) = self.createShareConfirmationViewController() else { return }
 
+        shareConfirmationVC.setChatMessage(self.textView.text)
+        self.setChatMessage("")
+
         picker.dismiss(animated: true) {
             self.present(navigationController, animated: true) {
                 for result in results {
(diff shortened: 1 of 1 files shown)


Ivansss (approved):
Really nice :) 
Thanks for this PR!

SystemKeeper (approved):
🥳

PR #2368, opened 2026-01-26
fix(reactions): Prevent accidental reaction toggle when long pressing

luflow:
## Summary

- Fixes accidental reaction toggles when users long press on a reaction to view "who reacted"
- Tracks touch duration to differentiate quick taps from long press attempts
- Only triggers reaction toggle for taps ≤ 0.25 seconds

## Problem

When long pressing on a reaction bubble to see "who reacted", users would sometimes accidentally toggle the reaction (add/remove it) if they released their finger before the context menu threshold (~0.5s) was reached.

### Root Cause

The `ReactionsView` (UICollectionView) handles taps via `didSelectItemAt` independently of the table view's context menu system. When a long press didn't quite reach the context menu threshold, the touch was interpreted as a tap, triggering an unwanted reaction toggle.

## Solution

Track the touch duration in `ReactionsView`:
- `touchesBegan` records the start time
- `touchesCancelled` clears it
- `didSelectItemAt` checks the duration:
  - **≤ 0.25s** → Quick tap → Toggle reaction ✓
  - **> 0.25s** → Long press attempt → Ignore (context menu handles it)

This threshold (0.25s) provides a buffer below the context menu's ~0.5s threshold while still allowing normal quick taps to work.

## Test plan

- [ ] Quick tap on a reaction toggles it (adds/removes your reaction)
- [ ] Long press on a reaction shows "who reacted" summary
- [ ] Releasing a long press early (before context menu appears) does NOT accidentally toggle the reaction
- [ ] Scrolling away from a reaction does not trigger any action

### DIFF
--- a/NextcloudTalk/Chat/Chat views/Reactions/ReactionsView.swift
+++ b/NextcloudTalk/Chat/Chat views/Reactions/ReactionsView.swift
@@ -14,6 +14,12 @@ import UIKit
     public weak var reactionsDelegate: ReactionsViewDelegate?
     var reactions: [NCChatReaction] = []
 
+    /// Tracks the touch start time to differentiate quick taps from long presses
+    private var touchBeganTime: Date?
+
+    /// Maximum duration (in seconds) for a touch to be considered a tap vs long press
+    private let maxTapDuration: TimeInterval = 0.25
+
     required init?(coder aDecoder: NSCoder) {
         super.init(coder: aDecoder)
         self.setupReactionView()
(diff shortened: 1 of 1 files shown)


Ivansss:
Hi @luflow :) Thank you for this PR!
Even though I experienced this issue in the past, I'm no longer able to reproduce it on a physical iOS device (iOS 26.2.1). Are you able to reproduce the issue reliably? Are you using a physical device or a simulator, and which iOS version are you testing with?

Ivansss (approved):

SystemKeeper (approved):
Thank you :)

PR #2392, opened 2026-02-09
feat: Allow password change in room settings

Ivansss:


SystemKeeper:
After changing the password, we are still in the "change password" state, should that be reset?

https://github.com/user-attachments/assets/01775cb7-d4ad-41dc-8c17-9f11651a3a6d

SystemKeeper (approved):

PR #2401, opened 2026-02-20
fix: Prevent re-joining room when leaving chat

Ivansss:
Steps to reproduce:

- Join one open conversation
- Go to "Conversation settings"
- Tap on "Leave conversation"

Without this PR:

User will rejoin the conversation after leaving it, because the chat view will rejoin the conversation (in `viewWillAppear()`) while moving to conversations list.

With this PR:
Prevents any action on chat view's `viewWillAppear()` if `leaveChat()` has been called.


SystemKeeper:
I guess we can't just reuse the `hasStopped` flag?

SystemKeeper (approved):
Thanks for adding the test 💛

PR #2404, opened 2026-02-25
feat: Chat relay

Ivansss:
- [x] Use chat relay when external signaling supports it
- [x] Fetch messages from server when websocket connection is lost (fetch messages looking into the future from last message we have in chat)
- [x] Set read marker for chat relay messages

### DIFF
--- a/NextcloudTalk/Chat/NCChatController.swift
+++ b/NextcloudTalk/Chat/NCChatController.swift
@@ -33,8 +33,11 @@ public class NCChatController: NSObject {
     private var getHistoryTask: URLSessionDataTask?
     private var pullMessagesTask: URLSessionDataTask?
 
-    private var chatRelayActive = false
-    private var chatRelayCatchUpInProgress = false
+    private enum ChatRelayState {
+        case inactive, active, catchingUp
+    }
+
+    private var chatRelayState: ChatRelayState = .inactive
     private var chatRelayMessagesBuffer: [[String: Any]] = []
     private var chatRelayMessagesQueue: DispatchQueue?
     private var externalSignalingController: NCExternalSignalingController?
@@ -433,7 +436,7 @@ public class NCChatController: NSObject {
               let messageDict = notification.userInfo?["message"] as? [String: Any] else { return }
 
         chatRelayMessagesQueue?.async {
-            if !self.chatRelayActive {
+            if self.chatRelayState != .active {
                 self.chatRelayMessagesBuffer.append(messageDict)
                 return
             }
@@ -443,8 +446,7 @@ public class NCChatController: NSObject {
 
     private func startProcessingChatRelayMessages() {
         chatRelayMessagesQueue?.async {
-            self.chatRelayCatchUpInProgress = false
-            self.chatRelayActive = true
+            self.chatRelayState = .active
             self.flushChatRelayMessagesBuffer()
         }
     }
@@ -467,10 +469,9 @@ public class NCChatController: NSObject {
     // payload (file/object shares, call_ended, unknown system messages). Incoming chat relay messages
     // are buffered while the catch-up runs.
     private func triggerChatRelayCatchUp() {
-        guard !chatRelayCatchUpInProgress else { return }
+        guard chatRelayState != .catchingUp else { return }
 
-        chatRelayCatchUpInProgress = true
-        chatRelayActive = false
+        chatRelayState = .catchingUp
 
         let lastChatBlock = chatBlocksForRoomOrThread().last
         let fromMessageId = lastChatBlock?.newestMessageId ?? 0
@@ -1038,8 +1039,7 @@ public class NCChatController: NSObject {
     }
 
     public func stopReceivingNewChatMessages() {
-        chatRelayActive = false
-        chatRelayCatchUpInProgress = false
+        chatRelayState = .inactive
         stopChatMessagesPoll = true
         pullMessagesTask?.cancel()
     }
--- a/NextcloudTalk/Chat/NCChatController.swift
+++ b/NextcloudTalk/Chat/NCChatController.swift
@@ -42,6 +42,9 @@ public class NCChatController: NSObject {
     private var chatRelayMessagesQueue: DispatchQueue?
     private var externalSignalingController: NCExternalSignalingController?
 
+    // Debounces the read-marker requests we issue while receiving messages over the chat relay. Only accessed on the main queue.
+    private var setReadMarkerWorkItem: DispatchWorkItem?
+
     public init!(for room: NCRoom) {
         guard let account = NCDatabaseManager.sharedInstance().talkAccount(forAccountId: room.accountId) else { return nil }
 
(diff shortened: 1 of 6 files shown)


NextcloudTalk/Chat/NCChatController.m
```diff
@@ -901,12 +933,59 @@ - (void)startReceivingChatMessagesFromMessagesId:(NSInteger)messageId withTimeou
         [self checkLastCommonReadMessage:lastCommonReadMessage];
         
         if (error.code != -999) {
+            BOOL shouldStartLongPolling = statusCode == 304;
             NCChatBlock *lastChatBlock = [self chatBlocksForRoomOrThread].lastObject;
-            [self startReceivingChatMessagesFromMessagesId:lastChatBlock.newestMessageId withTimeout:YES];
+
+            if (shouldStartLongPolling && self->_externalSignalingController.hasChatRelay) {
```
SystemKeeper: Wrong name of the variable?

NextcloudTalk/Chat/NCChatController.swift
```diff
@@ -408,6 +417,176 @@ public class NCChatController: NSObject {
...
+        guard let signalingController = NCSettingsController.sharedInstance().externalSignalingController(forAccountId: account.accountId),
+              signalingController.hasChatRelay else { return }
+        externalSignalingController = signalingController
+        chatRelayMessagesQueue = DispatchQueue(label: "chat.relay.message.queue")
+        NotificationCenter.default.addObserver(self, selector: #selector(didReceiveChatMessageFromExternalSignaling(_:)), name: .extSignalingDidReceiveChatMessage, object: signalingController)
+        NotificationCenter.default.addObserver(self, selector: #selector(didReconnectExternalSignaling(_:)), name: .extSignalingDidReconnect, object: signalingController)
+    }
+
+    @objc private func didReceiveChatMessageFromExternalSignaling(_ notification: Notification) {
+        guard let roomToken = notification.userInfo?["roomid"] as? String, roomToken == room.token,
```
SystemKeeper: In all other cases (if I did not miss one), we use `roomToken` outside of external signaling, so in external signaling, the notification userInfo contains `roomToken` as key, mapped from `roomid`. See `processRoomParticipantsEvents` for example

NextcloudTalk/Chat/NCChatController.swift
```diff
@@ -408,6 +417,176 @@ public class NCChatController: NSObject {
...
+              let chatDict = (notification.userInfo?["data"] as? [String: Any])?["chat"] as? [String: Any],
+              let messageDict = chatDict["comment"] as? [String: Any] else { return }
```
SystemKeeper: Generally, should we have that kind of processing not in controller, but in external signaling and pass on the processed information? The implementation detail of the signaling messages should be in external signaling, right?

NextcloudTalk/Chat/NCChatController.swift
```diff
@@ -408,6 +417,176 @@ public class NCChatController: NSObject {
...
+    }
+
+    // Falls back to a chat API catch-up when a chat relay message can't be reliably rendered from its
+    // payload (file/object shares, call_ended, unknown system messages). Incoming chat relay messages
+    // are buffered while the catch-up runs.
+    private func triggerChatRelayCatchUp() {
+        guard !chatRelayCatchUpInProgress else { return }
+
+        chatRelayCatchUpInProgress = true
+        chatRelayActive = false
```
SystemKeeper: Should we use an enum here, e.g. `chatRelayState`, with `.inactive, .active, .catchingUp` or something similar?

NextcloudTalk/Chat/NCChatController.swift
```diff
@@ -408,6 +417,176 @@ public class NCChatController: NSObject {
...
+
+        let lastChatBlock = chatBlocksForRoomOrThread().last
+        let fromMessageId = lastChatBlock?.newestMessageId ?? 0
```
SystemKeeper, line 480: I think we should guard this here, if `lastChatBlock?.newestMessageId` is nil, we fall back to 0 and then we would fetch _all_ messages from API.

NextcloudTalk/Chat/NCChatMessage.swift
```diff
@@ -49,6 +49,12 @@ import SwiftyAttributes
         return self.messageType == kMessageTypeVoiceMessage
     }
 
+    // Whether the message carries a file/media attachment (file, file2, …). Unlike `file()`, this
+    // is also true for messages with multiple file parameters.
+    public var hasFileParameter: Bool {
+        return self.messageParameters.keys.contains { ($0 as? String)?.hasPrefix("file") ?? false }
```
SystemKeeper: While it is true, that unlike `file()` this supports multiple files, this is currently not supported on server, so I would prefer to keep the `file() != nil` check

NextcloudTalk/Chat/NCChatController.swift
```diff
@@ -840,8 +1019,16 @@ public class NCChatController: NSObject {
...
                 let lastChatBlock = self.chatBlocksForRoomOrThread().last
-                self.startReceivingChatMessages(fromMessagesId: lastChatBlock?.newestMessageId ?? 0, withTimeout: true)
+
+                if shouldStartLongPolling, let extSignaling = self.externalSignalingController, extSignaling.hasChatRelay {
+                    print("Chat is up to date, now processing new messages from the chat relay")
+                    self.startProcessingChatRelayMessages()
+                    return
+                }
+
+                self.startReceivingChatMessages(fromMessagesId: lastChatBlock?.newestMessageId ?? 0, withTimeout: shouldStartLongPolling)
```
SystemKeeper: It's from the swift migration, but same issue as above, if we fallback to 0, we will get _all_ messages.

NextcloudTalk/Chat/NCSystemMessageLocalizer.swift
```diff
@@ -0,0 +1,168 @@
...
+                    : NSLocalizedString("{actor} demoted you from moderator", comment: "")
+            }
+            return cliIsActor(message)
+                ? NSLocalizedString("An administrator demoted {user} from moderator", comment: "")
+                : NSLocalizedString("{actor} demoted {user} from moderator", comment: "")
+
+        case "file_shared", "object_shared":
+            // Backend transforms these to normal chat messages, they should not be relayed as
+            // system messages. Fall back to the chat API to get the authoritative message.
+            return nil
```
SystemKeeper: I would move that to the same "block" as the other `return nil`

NextcloudTalk/Chat/NCSystemMessageLocalizer.swift
```diff
@@ -0,0 +1,168 @@
...
+        }
+    }
+
+    // MARK: - Helpers
+
+    // The message's actor is the local user. iOS accounts always authenticate as a regular user,
+    // so this matches the web's selfIsActor for the only actor type we can be.
+    private static func selfIsActor(_ message: NCChatMessage, _ account: TalkAccount) -> Bool {
+        return message.isMessage(from: account.userId)
+    }
```
SystemKeeper: Should that be part of `NCChatMessage`?

NextcloudTalk/Chat/NCSystemMessageLocalizer.swift
```diff
@@ -0,0 +1,168 @@
...
+    private static func selfIsUser(_ message: NCChatMessage, _ account: TalkAccount) -> Bool {
+        guard let userParameter = NCMessageParameter(dictionary: message.messageParameters["user"] as? [String: Any]) else {
+            return false
+        }
+        return userParameter.type == "user" && userParameter.parameterId == account.userId
+    }
+
+    private static func cliIsActor(_ message: NCChatMessage) -> Bool {
+        return message.actorId == "cli" && message.actorType == "guests"
+    }
```
SystemKeeper: Similar thing here, should it be part of `NCChatMessage` instead?

NextcloudTalk/Chat/NCChatController.m
Ivansss: Fixed in 86e1d5a954de1dbaf97849f6b5afe9b57865c256

NextcloudTalk/Chat/NCChatController.swift
Ivansss: Fixed in 86d9e9706080daa921e65a51dfcb554d5eaffca7

NextcloudTalk/Chat/NCChatController.swift
Ivansss: Fixed in 86d9e9706080daa921e65a51dfcb554d5eaffca7

NextcloudTalk/Chat/NCChatController.swift
Ivansss: Fixed in 138df80f2aa6c28931fe453fd1b5c0da885264d8

NextcloudTalk/Chat/NCChatMessage.swift
Ivansss: Fixed in 53c157e4f0cf572f2c39372346ff1855e2dcde2f

NextcloudTalk/Chat/NCSystemMessageLocalizer.swift
Ivansss: Fixed in d79b61d457ed455635c2c8155b165309d6cf8e08

NextcloudTalk/Chat/NCSystemMessageLocalizer.swift
Ivansss: Fixed in d79b61d457ed455635c2c8155b165309d6cf8e08

NextcloudTalk/Chat/NCSystemMessageLocalizer.swift
Ivansss: Fixed in d79b61d457ed455635c2c8155b165309d6cf8e08

NextcloudTalk/Chat/NCChatController.swift
Ivansss: Due to Swift migration, possible refactor in a follow up as discussed in private

NextcloudTalk/Chat/NCChatController.swift
Ivansss: Due to Swift migration, possible refactor in a follow up as discussed in private

SystemKeeper:
> *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Modifications to the layout engine must not be performed from a background thread after it has been accessed from the main thread.'
*** First throw call stack:
(0x1a25fa23c 0x19f0c9224 0x1f39bf3a4 0x1f39c1a1c 0x1a815d934 0x1a8131fa8 0x1a8258378 0x1a318eb94 0x1a318e6e4 0x1a2f80c88 0x1a2f3f110 0x1a2f6b61c 0x1a30540a0 0x2016dcc48 0x2016dc9b8 0x2016dde58 0x2016d9438 0x2016d88c0)
libc++abi: terminating due to uncaught exception of type NSException

NSNotifications are now dispatched on the chat relay queue instead of main and therefore crash, e.g. when starting a silent call on web, while in the room on ios.

image

Compare to main-branch:

image

SystemKeeper (approved):

PR #2405, opened 2026-02-25
fix: Only show pin option for moderators/owners

SystemKeeper:
Currently we show the pinning option for all participants, but we need moderator permission (or owner).


SystemKeeper:
Hmpf, need to trigger backport tomorrow.

Ivansss (approved):

PR #2420, opened 2026-03-09
fix: Set a name for the RoomsTableViewController

### LINKED ISSUE #2403: Add name to RoomsTableVC
* Enter conversation
* long press back button

See empty list entry

SystemKeeper:
* Fixes https://github.com/nextcloud/talk-ios/issues/2403

I was not able to reproduce the original issue on iOS 18 and 26 anymore. So it might be fixed in an earlier iOS version already, given that the change is from 2022 (https://github.com/nextcloud/talk-ios/pull/923)


SystemKeeper:
Ah probably now changed in tests... Will take a look.

Ivansss (approved):

Issue #916, opened 2022-10-12
Move to SPM


SystemKeeper:
As SPM is increasingly adopted, we should think about removing the dependency on CocoaPods and use SPM instead.

| Package | URL | SPM supported |
| --- | --- | --- |
| AFNetworking | https://github.com/AFNetworking/AFNetworking | ✅ |
| MaterialComponents/ActivityIndicator | https://github.com/material-components/material-components-ios | ⛔️ |
| MobileVLCKit | https://github.com/videolan/vlckit | ⛔️ |

marinofaggiana:
MBProgressHUD is very old ... advise JGProgressHUD
AFNetworking is full of bugs and too OLD (please replace it)
DateTools and MZTimerLabel ?

PR #2472, opened 2026-04-14
More SPM

SystemKeeper:
* Ref https://github.com/nextcloud/talk-ios/issues/916

---

* Move toast to SPM
* Remove usage of MZTimerLabel
* ~~Upgrade AFNetworking to 4.0.1 and switch to SPM~~ (postponed)


SystemKeeper:
Let's keep it like this and do everything else in a follow-up

Ivansss (approved):

PR #2477, opened 2026-04-16
fix: Open conversation test

### LINKED ISSUE #2459: Open conversation test fails on retry, when we joined already on the first try

Ivansss:
Check if user already joined to the open conversation (due to leave conversation failed)

Fix #2459


SystemKeeper (approved):
Tested and works. I am a bit sceptical on the 3s timeout, but let's find out :)

Issue #2460, opened 2026-04-12
API Controller migration follow-up


SystemKeeper:
- [x] Investigate https://github.com/nextcloud/talk-ios/pull/594
- [x] Investigate https://github.com/nextcloud/talk-ios/commit/147b34f2c8921eb16a3dc8dd750b8f7911c4f02b
- [x] Ensure for swift only methods we return OCSError instead of Error https://github.com/nextcloud/talk-ios/pull/2488

PR #2488, opened 2026-04-22
fix: Use correct API return types when possible

SystemKeeper:
* Ref https://github.com/nextcloud/talk-ios/issues/2460


SystemKeeper:
* Added https://github.com/nextcloud/talk-ios/pull/2488/commits/16ddadc28c979bae401c961978a6a8b3d2d772af to fix self signed certificates
* Rebased on main

Ivansss (approved):

PR #2492, opened 2026-04-29
feat: Permanent call rooms

Ivansss:
Ref: https://github.com/nextcloud/spreed/pull/17169

### DIFF
--- a/NextcloudTalk/Rooms/NCRoom.swift
+++ b/NextcloudTalk/Rooms/NCRoom.swift
@@ -372,4 +372,12 @@ import SwiftyAttributes
         return self.isUserOwnerOrModerator && NCDatabaseManager.sharedInstance().roomHasTalkCapability(kCapabilityPinnedMessages, for: self)
     }
 
+    // MARK: - Room Attributes
+
+    private static let voiceRoomAttributeBit: Int = 1
+
+    public var isVoiceRoom: Bool {
+        return (self.attributes & NCRoom.voiceRoomAttributeBit) != 0
+    }
+
 }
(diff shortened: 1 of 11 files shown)


NextcloudTalk/Rooms/NCRoom.swift
```diff
@@ -372,4 +372,12 @@ import SwiftyAttributes
         return self.isUserOwnerOrModerator && NCDatabaseManager.sharedInstance().roomHasTalkCapability(kCapabilityPinnedMessages, for: self)
     }
 
+    // MARK: - Room Attributes
+
+    private static let voiceRoomAttributeBit: Int = 1
+
+    public var isVoiceRoom: Bool {
+        return (self.attributes & NCRoom.voiceRoomAttributeBit) != 0
+    }
```
SystemKeeper: Can we use an enum/NS_OPTIONS here? E.g. https://github.com/nextcloud/talk-ios/blob/c8c66b4c46011e759fe9f44abfe004a43a042a80/NextcloudTalk/NCTypes.h#L18-L24

SystemKeeper (approved):
Tested and works, only nit pick question.

SystemKeeper:
One thing came to my mind, does this still work fine when entering via a push notification ?

Ivansss:
> One thing came to my mind, does this still work fine when entering via a push notification ?

Yes, that was the reason to add the logic in `startChat(inRoom room: NCRoom)` in `NCRoomsManager` :)

SystemKeeper:
Nice, thanks for the enum :)

PR #2506, opened 2026-05-07
fix: Do not use file names in temporary files uploaded to draft folder

### LINKED ISSUE #2502: Don't use filename when uploading with conversation folders
https://github.com/nextcloud/talk-ios/blob/388efa4e32c97101f5cbc5ae23991417b577750f/ShareExtension/ShareConfirmationViewController.swift#L659

Since we append the filename, we could run into the same issue that is described at https://github.com/nextcloud/spreed/issues/17896 ?
Or need to use the renamed Filename that the server returns.

Ivansss:
Format: `UUID.ext`

Fix #2502

### DIFF
--- a/NextcloudTalk/Chat/BaseChatViewController.swift
+++ b/NextcloudTalk/Chat/BaseChatViewController.swift
@@ -4116,7 +4116,9 @@ import Toast
                     failure("Probe conversation attachment folder returned no folder")
                     return
                 }
-                let tempName = UUID().uuidString + "." + (URL(string: fileName)?.pathExtension ?? "")
+                let fileExtension = URL(string: fileName)?.pathExtension ?? ""
+                let extensionSuffix = fileExtension.isEmpty ? "" : ".\(fileExtension)"
+                let tempName = UUID().uuidString + extensionSuffix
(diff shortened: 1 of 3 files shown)


NextcloudTalk/Chat/BaseChatViewController.swift
```diff
@@ -4116,7 +4116,7 @@ import Toast
                     failure("Probe conversation attachment folder returned no folder")
                     return
                 }
-                let tempName = UUID().uuidString + "-" + fileName
+                let tempName = UUID().uuidString + "." + (URL(string: fileName)?.pathExtension ?? "")
```
SystemKeeper: This might not be a big real-world issue, but in theory we can now end up with a dot at the end of the temporary filename, which is also not allowed on windows:

> Do not end a file or directory name with a space or a period. Although the underlying file system may support such names, the Windows shell and user interface does not. 

https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions

Shall we only add the dot if we have an extension?

Ivansss: Good point 👍 I will change it

SystemKeeper (approved):

PR #2517, opened 2026-05-11
fix: Use longpolling session manager for internal signaling

SystemKeeper:
Since the "normal" api session manager is limited to 25s, we always run into a (unnecessary) timeout.


SystemKeeper:
Well... Or not.

SystemKeeper:
> Subsequent calls to the this method with the same identifier will return the same storage instance.

Contrary to apples documentation, that seems not to be the case. It might be a side-effect of us not actually using a group container as an identifier, but an arbitrary string (accountId) to retrieve the storage, but we should make sure that there are not multiple session cookies for the same account.

Ivansss (approved):

PR #2526, opened 2026-06-01
fix(ChatViewController): Update icons for new guidelines

kra-mo:
See
https://developer.apple.com/design/human-interface-guidelines/toolbars#:~:text=Prefer%20system%2Dprovided%20symbols%20without%20borders


SystemKeeper:
Thank you! We will take a look after the release

Ivansss (approved):

PR #2561, opened 2026-06-20
chore: Migrate multiple classes to swift

SystemKeeper:
## 🤖 AI (if applicable)


SystemKeeper:
Also pushed a fix for threads (regression from https://github.com/nextcloud/talk-ios/pull/2556), see https://github.com/nextcloud/talk-ios/pull/2561/changes/8b250eba97432b38b15b8eb9cf41e423473427f1

NextcloudTalk/Database/NCDatabaseManager.swift
```diff
@@ -1,10 +1,769 @@
...
+
+    public func numberOfAccounts() -> Int {
+        return Int(TalkAccount.allObjects().count)
+    }
+
+    public func activeAccount() -> TalkAccount {
+        if let managedActiveAccount = TalkAccount.objects(where: "active = true").firstObject() as? TalkAccount {
+            return TalkAccount(value: managedActiveAccount)
+        }
+        return TalkAccount()
```
SystemKeeper, line 187: This can be tricky, but at the same time, if we change to an optional, this will be quite a big change (should be done separately I'd say?).
Would blame objc <> swift for now

NextcloudTalk/User Interface/NCUserInterfaceController.swift
```diff
@@ -0,0 +1,479 @@
...
+            if let serverURL {
+                loginViewController?.startLoginProcess(serverURL: serverURL, user: user)
+            }
+        }
+    }
+
+    func presentLoggedOutInvalidCredentialsAlert() {
+        let alert = UIAlertController(title: NSLocalizedString("Logged out", comment: ""),
+                                      message: NSLocalizedString("Credentials for this account were no longer valid", comment: ""),
+                                      preferredStyle: .alert)
```
SystemKeeper, line 85: Follow-up: Would make sense to show the user/server here as the request could be from another account in the background and not the active one. Similar to `presentAccountNotConfiguredAlert`

NextcloudTalk/User Interface/NCUserInterfaceController.swift
```diff
@@ -0,0 +1,479 @@
...
+    func presentShareLinkDialog(for room: NCRoom, inViewContoller viewController: UITableViewController?, for indexPath: IndexPath?) {
+        guard let roomLinkURL = room.linkURL else {
+            return
+        }
+
+        let items = [roomLinkURL]
+        let controller = UIActivityViewController(activityItems: items, applicationActivities: nil)
+
+        let appDisplayName = Bundle.main.infoDictionary?["CFBundleDisplayName"] as? String ?? ""
+        let emailSubject = String(format: NSLocalizedString("%@ invitation", comment: ""), appDisplayName)
```
SystemKeeper: Could we just use the name from `NCAppBranding` here? 🤔

NextcloudTalk/User Interface/NCUserInterfaceController.swift
```diff
@@ -0,0 +1,479 @@
...
+        }
+
+        controller.completionWithItemsHandler = { _, _, _, error in
+            if let error {
+                NSLog("An Error occured sharing room: %@, %@", error.localizedDescription, (error as NSError).localizedFailureReason ?? "")
+            }
+        }
+    }
+
+    func presentVoiceRoomJoinAlert(for room: NCRoom) {
```
SystemKeeper, line 357: This is very similar to `presentAlert(for pushNotification: NCPushNotification)`, maybe a followup makes sense?

NextcloudTalk/Database/NCDatabaseManager.swift
Ivansss: Yes, let's do it in a follow up PR since it will require quite some changes.

NextcloudTalk/User Interface/NCUserInterfaceController.swift
Ivansss: Follow up makes sense 👍

NextcloudTalk/User Interface/NCUserInterfaceController.swift
Ivansss: We could just `talkAppName` sure :)

NextcloudTalk/User Interface/NCUserInterfaceController.swift
Ivansss: We can try to merge them into one ;)

Ivansss (approved):

PR #2568, opened 2026-06-25
fix: Handle chat refresh requests from HPB

Ivansss:
fix: Handle chat refresh requests from the HPB (only if chat-relay is supported)

## 🤖 AI (if applicable)


SystemKeeper:
Does it work like that? From the documentation, if chat-relay is supported for the signaling server, we receive either a refresh or the full message? https://nextcloud-spreed-signaling.readthedocs.io/en/latest/standalone-signaling-api-v1/#send-chat-room-message

https://github.com/strukturag/nextcloud-spreed-signaling/blob/d82f7f2064f06e841fc7cddb97e6fa543a9df776/server/clientsession.go#L1438-L1444

and on older signaling systems we would receive both, the message and the refresh

SystemKeeper (approved):
Looks like I stand corrected :)

PR #2571, opened 2026-06-26
fix: Prevent duplicate peer crash in call collection view

Ivansss:
- [x] Prevent duplicate peer crash in call collection view
- [x] Stop sending current media state when leaving a call

## 🤖 AI (if applicable)

### DIFF
--- a/NextcloudTalk/Calls/CallViewController.swift
+++ b/NextcloudTalk/Calls/CallViewController.swift
@@ -1876,7 +1876,7 @@ class CallViewController: UIViewController,
             }
 
             for peerConnection in self.screenPeersInCall {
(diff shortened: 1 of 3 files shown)


NextcloudTalk/Calls/CallViewController.swift
```diff
@@ -1876,7 +1876,7 @@ class CallViewController: UIViewController,
             }
 
             for peerConnection in self.screenPeersInCall {
-                let screenRenderer = self.screenRenderersDict[peerConnection.peerIdentifier]
+                let screenRenderer = self.screenRenderersDict[peerConnection.peerId]
```
SystemKeeper, line 1879: Really good catch!

SystemKeeper (approved):

PR #2572, opened 2026-06-26
fix: Don't show read marker without message afterwards

SystemKeeper:
How to reproduce (all on ios)
* Create a thread
* Write a few messages outside of the thread
* Write a few messages inside of the thread
* Mark the conversation as unread with swiping the conversation for unread
* Enter the conversation
* See that the read marker is at the very bottom of the chat, without any message afterwards

## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI


SystemKeeper:
As discussed will either hide it (like web) or show a change the message to indicate that the message is in a thread

PR #2578, opened 2026-06-28
chore: Add RealmSwift lib

SystemKeeper:
* See also https://github.com/nextcloud-deps/realm-swift-binary/pull/2

Adding RealmSwift additionally, this allows us to remove the RLMSupport file. It is also the intended way having objc and swift in one project (/realm-swift/issues/8749).

With that PR it is _not really_ required (besides the RLMSupport removal), but it should make things easier later to switch.

## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI


SystemKeeper:
Closing this for now, current RealmSwift is not build with module stability, which means we need a built per xcode/compiler version. The main issue seems to be (judging from a search https://github.com/realm/realm-swift/issues?q=is%3Aissue%20state%3Aclosed%20module%20stability) that object autodiscover is not working, and we need to explictly set objectClasses. That wouldn't be an issue for us, so we could think about compiling it with module stability ourselves and try.

PR #2604, opened 2026-07-14
fix: Speaker view followups

SystemKeeper:
## 🤖 AI (if applicable)


SystemKeeper:
~~Need to check something.~~

Cannot set priority to 1000, since then the shrinking will not work again, so using 999 now.

Ivansss (approved):

PR #2621, opened 2026-07-17
feat: 🏷️ Conversation tags

### LINKED ISSUE #2505: Conversation folders/tags
what could https://github.com/nextcloud/spreed/pull/17497 look like on mobile?

@Ivansss and I discussed the possibility of chips vs dropdown/accordion style UI for these conversation categories/sections/folders. Initially I was insistent on a dropdown style but looking at other apps on iOS it seems like the ship style is much more common.

Dropdowns:
- Allows saving state of collapsed/expanded so the user can collapse everything they don't want to see always 
- Allows more space on the top so we can overload the chip bar with more filters
- searching is easier (search across all categories and expand if collapsed)

Chips:
- more common UI pattern (signal, whatsapp, iOS mail)
- allows focusing on only one category at a time
- search can be confusing (search only in selected category? all categories? unclear)

| Dropdowns| Chips|
|---|---|
| Image | Image |

n/a

**Additional context**
n/a

Ivansss:
* Fixes https://github.com/nextcloud/talk-ios/issues/2505

IMG_E0354

## 🤖 AI (if applicable)

### DIFF
--- a/NextcloudTalk/Rooms/RoomsTableViewController.swift
+++ b/NextcloudTalk/Rooms/RoomsTableViewController.swift
@@ -909,7 +909,11 @@ class RoomsTableViewController: UITableViewController, CCCertificateDelegate, UI
             let unreadRooms = taggedRooms.filter { $0.unreadMessages > 0 }
             let title = isFavoritesTag ? NSLocalizedString("Favorites", comment: "'Favorites' meaning 'Favorite conversations'") : tag.name
 
-            chips.append(TagFilterChip(id: tag.tagId, title: title, unreadCount: unreadRooms.count, hasUnreadMention: unreadRooms.contains { $0.hasUnreadMention }))
+            chips.append(TagFilterChip(id: tag.tagId,
+                                       title: title,
+                                       unreadCount: unreadRooms.count,
+                                       mentioned: unreadRooms.contains { self.isRoomMentioned($0) },
+                                       groupMentioned: unreadRooms.contains { self.isRoomGroupMentioned($0) }))
         }
 
         let account = NCDatabaseManager.sharedInstance().activeAccount()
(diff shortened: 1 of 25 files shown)


NextcloudTalk/Rooms/RoomsTableViewController.swift
```diff
@@ -1400,6 +1599,22 @@ class RoomsTableViewController: UITableViewController, CCCertificateDelegate, UI
...
+        return room.unreadMention || room.type == .oneToOne || room.type == .formerOneToOne
+    }
+
+    private func isRoomGroupMentioned(_ room: NCRoom) -> Bool {
+        if NCDatabaseManager.sharedInstance().serverHasTalkCapability(.directMentionFlag) {
+            return room.unreadMention && !room.unreadMentionDirect
+        }
+
+        return false
+    }
```
SystemKeeper: Could move both to NCRoom?

NextcloudTalk/en.lproj/Localizable.strings
```diff
@@ -818,6 +830,9 @@
 "Could not start group conversation" = "Could not start group conversation";
 
 /* No comment provided by engineer. */
+"Could not update tags" = "Could not update tags";
+
+/* Create a new tag */
 "Create" = "Create";
```
SystemKeeper: "Create" already exists -> shall we unify the comment?

SystemKeeper (commented):
* It's weird that I can add tags in two places. IMO it should only be possible inside manage tags
* The checkmark for the tags in RoomInfo is very thin
* It feels a bit weird, that the thread tab now opens a modal, but nothing for this PR
* Downside of having archived as a pill: You don't easily spot the "dot" there anymore (is it an issue?)
* I don't like that the tags are only retrieved from the database on appReady. It takes a second to show them after app start, while the rooms are already there. Feels laggy
* Personally I don't see why we need a favorites tag 😓

SystemKeeper:
Do need to make the bar a tad bigger or a tad more padding to the conversations?

NextcloudTalk/Rooms/RoomsTableViewController.swift
Ivansss: Moved ;)

SystemKeeper (approved):

NextcloudTalk/en.lproj/Localizable.strings
Ivansss: Updated the comment

PR #2642, opened 2026-07-31
refactor(chat): Upload shared files with the single implementation

Ivansss:
Fourth and last refactoring step, on top of #2641. This is the one that removes the second upload
implementation: `ShareConfirmationViewController` had its own copy because `ChatFileUploader` was
not part of the ShareExtension target.

* Adds a batch `upload()` to the uploader: one draft folder request for the whole batch, uploads
  running concurrently, one result per file
* Uses it in `ShareConfirmationViewController` and deletes `startUploads()`, `uploadFile()` and the
  dispatch group and error/success lists around them
* Fixes a leftover send button in the navigation bar after sharing, which showed up next to the one
  in the input bar

Quota and rate limit failures now show the strings the app has for them instead of the ones from
NextcloudKit.

## 🤖 AI (if applicable)


NextcloudTalk/Chat/Chat upload/ChatFileUploader.swift
```diff
@@ -20,39 +20,99 @@ enum ChatFileUploader {
...
+                        let destination: ChatFileUploadDestination
+
+                        if let draftFolder {
+                            destination = try await self.draftFolderDestination(in: draftFolder, for: upload)
+                        } else {
+                            destination = try await self.resolveDestination(for: upload)
+                        }
+
+                        try await self.put(upload, to: destination, progress: { progress?(index, $0) }, mayCreateAttachmentFolder: true)
+                        try await self.announce(upload, at: destination)
```
SystemKeeper, line 59: One thing we should try to do in a followup is to retain the order of the shares.

ShareExtension/ShareConfirmationViewController.swift
```diff
@@ -622,152 +619,93 @@ import MBProgressHUD
...
+        case ChatFileUploadError.attachmentFolderUnavailable:
+            return NSLocalizedString("Failed to check or create attachment folder", comment: "")
+        case ChatFileUploadError.quotaExceeded:
+            return NSLocalizedString("User storage quota exceeded", comment: "")
+        case ChatFileUploadError.tooManyRequests:
+            return NSLocalizedString("Too many requests, please try again later", comment: "")
+        case ChatFileUploadError.uploadFailed(_, let errorDescription):
+            return errorDescription
+        default:
+            return error.localizedDescription
```
SystemKeeper, line 708: Very similar to the one in BsaeChatTableViewcontroller. Might make sense to check if we can unify them.

SystemKeeper (approved):

PR #2659, opened 2026-08-19
fix: Wrap updateConversationTags transaction in a background task

Ivansss:
Every realm write should run inside a background assertion, so iOS does not
suspend us while the realm write lock is held. The realm lives in the shared
app group container, so being suspended holding its lock is a kill. That was
open-coded at every call site, and about half the write sites were missing it.

`RLMRealm.writeTransaction` now does it in one place, releasing the assertion
with a `defer`, and `NCDatabaseManager.updateTalkAccount(forAccountId:)` covers
the common "change one field on the account" case.

`NCRoomsManager.updateRooms` is left as it is: it loops over every room on the
account, which can be hundreds, each with nested deletes. It is the only write
that could realistically run out of background time and need to bail out mid
transaction, so it keeps its own background task and `cancelWriteTransaction()`.
Everywhere else the block is a few statements, `updateConversationTags`
included — which is why its per item expiration checks are gone.

## 🤖 AI (if applicable)

### DIFF
--- a/NextcloudTalk/Database/NCDatabaseManager.swift
+++ b/NextcloudTalk/Database/NCDatabaseManager.swift
@@ -886,17 +886,30 @@ public extension Notification.Name {
     }
 
     func updateConversationTags(_ tags: [NCConversationTag], forAccountId accountId: String) {
+        let bgTask = BGTaskHelper.startBackgroundTask(withName: "updateConversationTagsForAccountId", expirationHandler: nil)
         let realm = RLMRealm.default()
 
         try? realm.transaction {
+            if bgTask.isExpired {
+                realm.cancelWriteTransaction()
+                return
+            }
+
             let query = NSPredicate(format: "accountId = %@", accountId)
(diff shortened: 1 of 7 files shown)


NextcloudTalk/Database/NCDatabaseManager.swift
```diff
@@ -886,17 +886,30 @@ public extension Notification.Name {
...
 
     func updateConversationTags(_ tags: [NCConversationTag], forAccountId accountId: String) {
+        let bgTask = BGTaskHelper.startBackgroundTask(withName: "updateConversationTagsForAccountId", expirationHandler: nil)
         let realm = RLMRealm.default()
 
         try? realm.transaction {
+            if bgTask.isExpired {
+                realm.cancelWriteTransaction()
+                return
+            }
```
SystemKeeper, line 896: As discussed, should probably be moved to outside of transaction

SystemKeeper (approved):

PR #2676, opened 2026-08-28
fix: Show all reactions of a message and keep their order stable

Ivansss:
### Fixes

- Reactions were clipped to the width the reused cell showed before, so a message
  could show only some of them — the rest were reachable by scrolling, with nothing
  indicating they were there.
- Reactions were sorted on top of an unordered dictionary, so the row came out in a
  different order on every launch and reactions moved even when their count hadn't
  changed. They now keep the order they were first used in, like the web client.

### Additions

- The row fades out at the edge when there are more reactions than fit.
- Haptic feedback when tapping a reaction.

## 🤖 AI (if applicable)

### DIFF
--- a/NextcloudTalk/Chat/Chat views/Reactions/ReactionsView.swift
+++ b/NextcloudTalk/Chat/Chat views/Reactions/ReactionsView.swift
@@ -66,6 +66,58 @@ import UIKit
         self.setContentOffset(.zero, animated: false)
     }
 
+    // MARK: - Scroll fade
+
+    /// Width of the fade shown at an edge that has more reactions behind it
+    private static let scrollFadeWidth: CGFloat = 16
+
+    private var scrollFadeLayer: CAGradientLayer?
+
+    override func layoutSubviews() {
+        super.layoutSubviews()
+
+        // Also called while scrolling, so the fade follows the content offset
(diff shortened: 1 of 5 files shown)


NextcloudTalk/Chat/Chat views/Reactions/ReactionsView.swift
```diff
@@ -66,6 +66,58 @@ import UIKit
...
+        let fadeLayer: CAGradientLayer
+        if let scrollFadeLayer {
+            fadeLayer = scrollFadeLayer
+        } else {
+            fadeLayer = CAGradientLayer()
+            fadeLayer.startPoint = .init(x: 0, y: 0.5)
+            fadeLayer.endPoint = .init(x: 1, y: 0.5)
+            self.scrollFadeLayer = fadeLayer
+            self.layer.mask = fadeLayer
+        }
```
SystemKeeper: I guess that could be just a lazy var?

SystemKeeper (approved):

NextcloudTalk/Chat/Chat views/Reactions/ReactionsView.swift
Ivansss: added ;)

PR #2688, opened 2026-09-02
feat: Show owner and moderator icons

Ivansss:
Replaces the "(moderator)" name suffix with an icon after the name.
Owners get a crown, moderators and guest moderators a shield. 
Skipped in conversations without ranks: one-to-one, former one-to-one, changelog and note-to-self.

## 🤖 AI (if applicable)


SystemKeeper (approved):
Tested and works.

* Any reason to use a view instead of an attachment to an attributed string?
* We should think about still showing a context menu, that shows the explanation of owner/moderator.

PR #2689, opened 2026-09-03
fix: Only hand polling over to the chat relay once the room is joined on the HPB

Ivansss:
The chat relay was armed on the chat API's 304 alone, without checking that our signaling session had
joined the room on the HPB. So we stopped polling before the HPB was sending us anything, and any
message sent in between was never received.

The relay now waits for the HPB room ack (new `joinedRoomToken`) and keeps long polling until then.

## 🤖 AI (if applicable)

### DIFF
--- a/NextcloudTalk/Chat/NCChatController.swift
+++ b/NextcloudTalk/Chat/NCChatController.swift
@@ -34,7 +34,7 @@ public class NCChatController: NSObject {
     private var getHistoryTask: URLSessionDataTask?
     private var pullMessagesTask: URLSessionDataTask?
 
-    private enum ChatRelayState {
+    enum ChatRelayState {
         case inactive, active, catchingUp
     }
 
@@ -58,7 +58,9 @@ public class NCChatController: NSObject {
 
         super.init()
 
-        setupChatRelay()
+        let signalingController = NCSettingsController.sharedInstance().externalSignalingController(forAccountId: account.accountId)
+        setupChatRelay(with: signalingController)
+
         AllocationTracker.shared.addAllocation("NCChatController")
     }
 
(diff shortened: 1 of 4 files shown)


NextcloudTalk/WebRTC/NCExternalSignalingController.swift
```diff
@@ -340,6 +346,15 @@ public enum NCExternalSignalingSendMessageStatus {
         let sessionChanged = self.sessionId != newSessionId
         self.sessionId = newSessionId
 
+        if sessionChanged {
+            // The new session did not join any room yet, the re-join below takes care of that
+            self.joinedRoomToken = nil
```
SystemKeeper, line 349: i wonder if we should always use the `setJoinedRoomToken` method?

NextcloudTalk/Chat/NCChatController.swift
```diff
@@ -463,6 +485,63 @@ public class NCChatController: NSObject {
...
+            guard self.chatRelayState == .waitingForJoin else { return }
+
+            // We are not waiting for the join anymore, we are waiting for the poll below to confirm
+            // that the chat is up to date. Also keeps a second join from polling twice.
+            self.chatRelayState = .inactive
+
+            DispatchQueue.main.async {
+                let lastChatBlock = self.chatBlocksForRoomOrThread().last
+                self.startReceivingChatMessages(fromMessagesId: lastChatBlock?.newestMessageId ?? 0, withTimeout: false)
+            }
```
SystemKeeper: This is a bit problematic. We now do
* 1 Request if we are quick enough with the signaling joining
* 2 Requests if we are quick enough with the signaling joining, but not yet started long polling
* 3 Requests if we are too slow, so we do an initial GET without timeout, one with timeout and then the final here.

NextcloudTalk/WebRTC/NCExternalSignalingController.swift
Ivansss: `setJoinedRoomToken` was removed since ExternalSignalingController doesn't notify anymore about joined room

NextcloudTalk/Chat/NCChatController.swift
Ivansss: Removed code in the new approach

SystemKeeper (approved):

PR #2705, opened 2026-09-10
feat(chat): Group the files of one upload into a single message

### LINKED ISSUE #2701: Multiple files sharing support 🖼️+🖼️
* See https://github.com/nextcloud/spreed/issues/11411

Ivansss:
IMG_E0377

Fixes https://github.com/nextcloud/talk-ios/issues/2701

## 🤖 AI (if applicable)

### DIFF
--- a/NextcloudTalk/Chat/Chat cells/BaseChatTableViewCell+FileGroup.swift
+++ b/NextcloudTalk/Chat/Chat cells/BaseChatTableViewCell+FileGroup.swift
@@ -1,0 +1,86 @@
+//
+// SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+// SPDX-License-Identifier: GPL-3.0-or-later
+//
+
+import UIKit
+
+extension BaseChatTableViewCell {
+
+    func setupForFileGroupCell(with group: FileMessageGroup, with account: TalkAccount) {
+        if self.groupedFilePreviewView == nil {
+            let groupedFilePreviewView = GroupedFilePreviewView()
+            self.groupedFilePreviewView = groupedFilePreviewView
+
+            groupedFilePreviewView.translatesAutoresizingMaskIntoConstraints = false
+            groupedFilePreviewView.delegate = self
+            groupedFilePreviewView.accessibilityIdentifier = "groupedFilePreviewView"
+
+            self.messageBodyView.addSubview(groupedFilePreviewView)
+
+            let messageTextView = MessageBodyTextView()
+            self.messageTextView = messageTextView
+
+            messageTextView.translatesAutoresizingMaskIntoConstraints = false
+
+            self.messageBodyView.addSubview(messageTextView)
+
+            // Without a caption the text view is taken out of the layout entirely. Hiding it is not
+            // enough: an empty text view still takes its line height, which would make the body
+            // taller than the cell was measured to be, and everything below it stops taking taps.
+            self.fileGroupCaptionConstraints = [
+                messageTextView.topAnchor.constraint(equalTo: groupedFilePreviewView.bottomAnchor, constant: 10),
+                messageTextView.bottomAnchor.constraint(equalTo: self.messageBodyView.bottomAnchor)
+            ]
+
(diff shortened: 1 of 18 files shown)


NextcloudTalk/Chat/Chat cells/BaseChatTableViewCell+FileGroup.swift
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        guard let groupedFilePreviewView = self.groupedFilePreviewView,
+              let messageTextView = self.messageTextView
+        else { return }
+
+        groupedFilePreviewView.setup(with: group, account: account, availableWidth: self.availableBodyWidth)
+
+        // The caption is carried by the file shared last, which is what the group is shown as
+        let hasCaption = !group.anchor.sharesFileWithoutCaption
```
SystemKeeper, line 54: I wonder if we can rely on that entirely? It’s up to the uploading client, isn’t it?

What happens with old clients? What happens with reaction on individual messages or captions (which is possible on old clients and api in general).

NextcloudTalk/Chat/Chat cells/GroupedFilePreviewTileView.swift
```diff
@@ -0,0 +1,310 @@
...
+// SPDX-License-Identifier: GPL-3.0-or-later
+//
+
+import UIKit
+
+extension NCMessageFileParameter {
+
+    /// Whether the file is shown as a preview tile rather than as a row with its name.
+    ///
+    /// Only media the server can render a preview of: everything else would be a tile showing the
```
SystemKeeper, line 12: That is not true, but might still be intended. But the server is able to generate previews for pdfs or office documents if setup correctly

NextcloudTalk/Chat/Chat upload/FileUploadReference.swift
```diff
@@ -0,0 +1,78 @@
...
+
+    /// Position of the file within its upload, starting at 1.
+    let position: Int
+
+    private static let uploadHashLength = 60
+    private static let positionDigits = 3
+
+    /// The characters the agreed format allows, which is stricter than what a hash could contain:
+    /// a reference id with an uppercase letter in it is not one of ours.
+    private static let hexDigits = Set("0123456789abcdef")
```
SystemKeeper: Why?

SystemKeeper (commented):
Just a first look at the first files. 

Do we handle edit and delete of individual messages correctly?
I generally find this PR very verbose comment wise ..

NextcloudTalk/Chat/Chat cells/BaseChatTableViewCell+FileGroup.swift
Ivansss: > I wonder if we can rely on that entirely? It’s up to the uploading client, isn’t it?

It is up to the client yes, but if the client is sending the group reference id (supporting grouping feature) I think we can assume the caption will be on the last message.

> What happens with old clients? What happens with reaction on individual messages or captions (which is possible on old clients and api in general).

This is a current issue, at the moment only the reactions of the last message of the group will be shown.
https://github.com/nextcloud/spreed/issues/19371

NextcloudTalk/Chat/Chat cells/GroupedFilePreviewTileView.swift
Ivansss: That's right, the comment needs to be updated but the functionality is intended (only images and video previews will be shown in the preview tiles)

NextcloudTalk/Chat/Chat upload/FileUploadReference.swift
Ivansss: The reference id format is defined as `/[a-f0-9]{60}-[0-9]{3}/` in https://github.com/nextcloud/spreed/pull/19040
Do you mean why implementing it like this and not with a regex?
I would update the comment since it's a bit confusing.

NextcloudTalk/Chat/Chat cells/BaseChatTableViewCell+FileGroup.swift
SystemKeeper: > I think we can assume the caption will be on the last message.

But if the API/old clients can edit message in between, that's still an issue?

